PR #21, opened 2019-05-18
Fix non isosceles triangle case for right_triangle with ORIENT_Z

ether42:
Hi there,

The first commit should fix the broken `ORIENT_Z` parameter for `right_triangle` when the triangle is not isosceles (when `width` isn't equal to `thickness`).
The second commit remove some redundancy, but feel free to discard it if you don't like it.

Test code: `right_triangle([60, 10, 40], orient=ORIENT_Z) frame_ref();`
Before: before
After:  after

Thanks for the library.


revarbat:
Thanks!

PR #44, opened 2019-06-12
Add strings, structures, supershape fix, list_set

adrianVmariano:
I'm a bit confused about why this says it can't merge and that I'm still behind after I updated.  But hopefully this works.  The changes include strings.scad, structs.scad, the supershapes fix in shapes2d and list_set, which is a dependency of the struct code.


revarbat:
You should be able to merge my master branch to your local copy, fix the conflict, commit it, then push it to your github fork.  Then the PR should lose the conflict.  I was going to try to fix it on the website, but their conflict resolution page seems broken.

PR #74, opened 2019-07-16
Added repeat to turtle.  Added centroid to geometry.scad.

adrianVmariano:


shapes2d.scad
```diff
@@ -1,3 +1,4 @@
+include<BOSL2/std.scad>
```
revarbat: This include line should be removed for the final submission

adrianVmariano:
Oops.  I fixed that.

PR #110, opened 2019-11-13
Corrected processing of $fn for round_corners, added two examples to skin (from list comp demos) and added another error check to zip.

adrianVmariano:


rounding.scad
```diff
@@ -327,6 +327,7 @@ function _bezcorner(points, parm) =
 function _circlecorner(points, parm) =
 	let(
 		angle = vector_angle(points)/2,
+                df=echo(angle=angle),
```
revarbat: Should remove debugging echos from main code base.

PR #140, opened 2020-03-15
Added xymove to turtle.

adrianVmariano:


math.scad
```diff
@@ -106,7 +106,9 @@ function factorial(n,d=1) = product([for (i=[n:-1:d]) i]);
 //   // Points colored in ROYGBIV order.
 //   rainbow(pts) translate($item) circle(d=3,$fn=8);
 function lerp(a,b,u) =
+        assert(same_shape(a,b), "Bad or inconsistent inputs to lerp")
 	is_num(u)? (1-u)*a + u*b :
+        assert(is_vector(u), "Input u to lerp must be number or vector")
```
revarbat: Actually, `lerp()` can take numbers, vectors, or ranges.  (Like `[3:2:8]`)

revarbat: I'll fix it in post.

PR #172, opened 2020-05-27
pie_slice can't accept `d` parameter

matthewfallshaw:
Since `r=10` is supplied as a default parameter in the module argument list, a call supplying a `d` parameter (eg. `pie_slice(d=10,h=10,ang=90);`) will error on the call to `get_radius` on line 1428 with `ERROR: Assertion 'is_undef(d)' failed: "Conflicting or redundant radius/diameter arguments given."`.
Since the calls to `get_radius` on lines 1428 and 1429 both supply `dflt=10`, line 1422 can just change to supplying `r=undef`.


revarbat:
Good catch.

PR #201, opened 2020-07-11
Updates

adrianVmariano:


tests/test_math.scad
```diff
@@ -1,586 +1,650 @@
-include <BOSL2/std.scad>
+include <BOSL/constants.scad>
```
revarbat, line 1: This is a BOSL1 include, not a BOSL2 include.

adrianVmariano: I copied the wrong file into my fork.  I have updated the file to the right version.

PR #205, opened 2020-07-17
input data checks and addition of many new functions, test_quaternions updated

RonaldoCMP:
1. Input data check in all functions
2. Definition of new functions for data check, new interpolations functions, quaternion from matrix, rotation path interpolating quaternions, inverse, exp, ln, power, and more.
3. test_quaternions were updated accordingly.

### DIFF
--- a/quaternions.scad
+++ b/quaternions.scad
@@ -319,7 +319,7 @@ function Q_Dist(q1, q2) =
 //   #sphere(r=80);
 function Q_Slerp(q1, q2, u, _dot) =
   is_undef(_dot)
-  ? assert(is_num(u) || is_num(0*u*u), "Invalid interpolation coefficient(s)")
+  ? assert(is_num(u) || is_range(u) || is_num(0*u*u), "Invalid interpolation coefficient(s)")
     assert(Q_is_quat(q1) && Q_is_quat(q2), "Invalid quaternion(s)" )
     let(
       _dot  = q1*q2,
@@ -522,7 +522,6 @@ function Q_Rotation_path(q1, n=1, q2) =
   ? [for( i=0, dR=Q_Matrix4(q1), R=dR; i<=n; i=i+1, R=dR*R ) R]
   : let( q2 = Q_Normalize( q1*q2<0 ? -q2: q2 ) )
     let( dq = Q_pow( Q_Mul( q2, Q_Inverse(q1) ), 1/n ),
-//    let( dq = Q_Mul( Q_Slerp(q1,q2, 1/n ), Q_Inverse(q1) ),
          dR = Q_Matrix4(dq) )
     [for( i=0, R=Q_Matrix4(q1); i<=n; i=i+1, R=dR*R ) R];
 
@@ -562,7 +561,7 @@ module Q_Rotation_path(q1, n=1, q2) {
 function Q_Nlerp(q1,q2,u) =
   assert(Q_is_quat(q1) && Q_is_quat(q2), "Invalid quaternion(s)" )
   assert( ! approx(norm(q1+q2),0), "Quaternions cannot be opposed" )
-  assert(is_num(0*u) || (is_list(u) && is_num(0*u*u)) ,
+  assert(is_num(0*u) || is_range(u) || (is_list(u) && is_num(0*u*u)) ,
          "Invalid interpolation coefficient(s)" )
   let( q1  = Q_Normalize(q1),
        q2  = Q_Normalize(q2) )
@@ -611,7 +610,11 @@ function Q_Nlerp(q1,q2,u) =
 //       Qrot(q) right(80) cube([10,10,1]);
 //   #sphere(r=80);
 function Q_Squad(q1,q2,q3,q4,u) =
-  Q_Slerp( Q_Slerp(q1,q4,u), Q_Slerp(q2,q3,u), 2*u*(1-u));
+  assert(is_num(0*u) || is_range(u) || (is_list(u) && is_num(0*u*u)) ,
+         "Invalid interpolation coefficient(s)" )
+	is_num(u)
+  ? Q_Slerp( Q_Slerp(q1,q4,u), Q_Slerp(q2,q3,u), 2*u*(1-u))
+	: Q_Slerp( Q_Slerp(q1,q4,u), Q_Slerp(q2,q3,u), [for(ui=u) 2*ui*(1-ui)]);
 
 
 // Function: Q_exp()
--- a/quaternions.scad
+++ b/quaternions.scad
@@ -47,11 +47,11 @@ function Q_is_quat(q) = is_vector(q,4) && ! approx(norm(q),0) ;
 //   ax = Vector of axis of rotation.
 //   ang = Number of degrees to rotate around the axis counter-clockwise, when facing the origin.
 function Quat(ax=[0,0,1], ang=0) = 
-  assert( is_vector(ax,3) && is_num(0*ang), "Invalid input")
-  let( n = norm(ax) )
-  approx(n,0) 
-  ? _Quat([0,0,0], sin(ang/2), cos(ang/2))
-  : _Quat(ax/n, sin(ang/2), cos(ang/2));
+    assert( is_vector(ax,3) && is_finite(ang), "Invalid input")
+    let( n = norm(ax) )
+    approx(n,0) 
+    ? _Quat([0,0,0], sin(ang/2), cos(ang/2))
+    : _Quat(ax/n, sin(ang/2), cos(ang/2));
 
 
 // Function: QuatX()
@@ -61,8 +61,8 @@ function Quat(ax=[0,0,1], ang=0) =
 // Arguments:
 //   a = Number of degrees to rotate around the axis counter-clockwise, when facing the origin.
 function QuatX(a=0) = 
-  assert( is_num(0*a), "Invalid angle" )
-  Quat([1,0,0],a);
+    assert( is_finite(a), "Invalid angle" )
+    Quat([1,0,0],a);
 
 
 // Function: QuatY()
@@ -72,8 +72,8 @@ function QuatX(a=0) =
 // Arguments:
 //   a = Number of degrees to rotate around the axis counter-clockwise, when facing the origin.
 function QuatY(a=0) =  
-  assert( is_num(0*a), "Invalid angle" )
-  Quat([0,1,0],a);
+    assert( is_finite(a), "Invalid angle" )
+    Quat([0,1,0],a);
 
 
 // Function: QuatZ()
@@ -83,8 +83,8 @@ function QuatY(a=0) =
 // Arguments:
 //   a = Number of degrees to rotate around the axis counter-clockwise, when facing the origin.
 function QuatZ(a=0) =  
-  assert( is_num(0*a), "Invalid angle" )
-  Quat([0,0,1],a);
+    assert( is_finite(a), "Invalid angle" )
+    Quat([0,0,1],a);
 
 
 // Function: QuatXYZ()
@@ -95,11 +95,11 @@ function QuatZ(a=0) =
 // Arguments:
 //   a = The triplet of rotation angles, [X,Y,Z]
 function QuatXYZ(a=[0,0,0]) =
-  assert( is_vector(a,3), "Invalid angles")
+    assert( is_vector(a,3), "Invalid angles")
     let(
-        qx = QuatX(a[0]),
-        qy = QuatY(a[1]),
-        qz = QuatZ(a[2])
+      qx = QuatX(a[0]),
+      qy = QuatY(a[1]),
+      qz = QuatZ(a[2])
     )
     Q_Mul(qz, Q_Mul(qy, qx));
 
@@ -111,14 +111,14 @@ function QuatXYZ(a=[0,0,0]) =
 //   Returns the normalized quaternion that rotates the non zero 3D vector v1 
 //   to the non zero 3D vector v2.
 function Q_From_to(v1, v2) =
-  assert( is_vector(v1,3) && is_vector(v2,3) 
-          && ! approx(norm(v1),0) && ! approx(norm(v2),0)
-          , "Invalid vector(s)")
-  let( ax = cross(v1,v2),
-       n  = norm(ax) )
-  approx(n, 0)
-  ? v1*v2>0 ? Q_Ident() : Quat([ v1.y, -v1.x, 0], 180)  
-  : Quat(ax, atan2( n , v1*v2 ));
+    assert( is_vector(v1,3) && is_vector(v2,3) 
+            && ! approx(norm(v1),0) && ! approx(norm(v2),0)
+            , "Invalid vector(s)")
+    let( ax = cross(v1,v2),
+         n  = norm(ax) )
+    approx(n, 0)
+    ? v1*v2>0 ? Q_Ident() : Quat([ v1.y, -v1.x, 0], 180)  
+    : Quat(ax, atan2( n , v1*v2 ));
 
 
 // Function: Q_Ident()
@@ -133,8 +133,8 @@ function Q_Ident() = [0, 0, 0, 1];
 //   Adds a scalar value `s` to the W part of a quaternion `q`.
 //   The returned quaternion is usually not normalized.
 function Q_Add_S(q, s) =  
-  assert( is_num(0*s), "Invalid scalar" )
-  q+[0,0,0,s];
+    assert( is_finite(s), "Invalid scalar" )
+    q+[0,0,0,s];
 
 
 // Function: Q_Sub_S()
@@ -144,8 +144,8 @@ function Q_Add_S(q, s) =
 //   Subtracts a scalar value `s` from the W part of a quaternion `q`.
 //   The returned quaternion is usually not normalized.
 function Q_Sub_S(q, s) =  
-  assert( is_num(0*s), "Invalid scalar" )
-  q-[0,0,0,s];
+    assert( is_finite(s), "Invalid scalar" )
+    q-[0,0,0,s];
 
 
 // Function: Q_Mul_S()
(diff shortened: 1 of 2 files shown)


quaternions.scad
```diff
@@ -20,51 +20,82 @@
...
 // Arguments:
 //   ax = Vector of axis of rotation.
 //   ang = Number of degrees to rotate around the axis counter-clockwise, when facing the origin.
-function Quat(ax=[0,0,1], ang=0) = _Quat(ax/norm(ax), sin(ang/2), cos(ang/2));
+function Quat(ax=[0,0,1], ang=0) = 
+  assert( is_vector(ax,3) && is_num(0*ang), "Invalid input")
+  let( n = norm(ax) )
+  approx(n,0) 
+  ? _Quat([0,0,0], sin(ang/2), cos(ang/2))
+  : _Quat(ax/n, sin(ang/2), cos(ang/2));
```
revarbat, line 54: Very minor formatting quibble for future reference... In most the library, the preferred multi-line formatting for ternary operators has the `?` at the end of the test expression, on the same line, and the `:` is usually at the end of the truth result expression.  The following are common formatting idioms for ternary ops, but are not enforced:

```openscad
a = test(x)? short_result_expr1 : short_result_expr2;

b = test(x)? somewhat_longer_result_expr1 :
    somewhat_longer_result_expr2;

c = test1(x)? short_result_expr1 : 
    test2(x)? short_result_expr2 : 
    test3(x)? short_result_expr3 : 
    default_result_expr;

d = test(x)? let(
    // complicated multi-line calculations here
) result1 : let(
    // complicated multi-line calculations here
) result2;

e = test(x)? (
    // more complicated subexpressions here, possibly with multiple let() blocks.
    result1
) : (
    // more complicated subexpressions here, possibly with multiple let() blocks.
    result2
);
```

Having said all that, I'll still accept this submitted code formatting.

quaternions.scad
```diff
@@ -20,51 +20,82 @@
...
 // Function: QuatX()
 // Usage:
 //   QuatX(a);
-// Description: Create a new Quaternion for rotating around the X axis [1,0,0].
+// Description: Create a normalized Quaternion for rotating around the X axis [1,0,0].
 // Arguments:
 //   a = Number of degrees to rotate around the axis counter-clockwise, when facing the origin.
-function QuatX(a=0) = Quat([1,0,0],a);
+function QuatX(a=0) = 
+  assert( is_num(0*a), "Invalid angle" )
```
revarbat, line 64: Is the `is_num(0*a)` instead of the equivalent `is_finite(a)` for speed?

quaternions.scad
```diff
@@ -73,6 +104,23 @@ function QuatXYZ(a=[0,0,0]) =
...
 
+// Function: Q_From_to()
+// Usage:
+//    q = Q_From_to(v1, v2);
+// Description: 
+//   Returns the normalized quaternion that rotates the non zero 3D vector v1 
+//   to the non zero 3D vector v2.
+function Q_From_to(v1, v2) =
+  assert( is_vector(v1,3) && is_vector(v2,3) 
+          && ! approx(norm(v1),0) && ! approx(norm(v2),0)
```
revarbat, line 115: The `!approx(norm(v),0)` idiom seems to occur frequently.  Do you think we should have a `null_vec(x)` function?  `vzero(x)`?  `is_zero(x)`?  `is_null(x)`?

quaternions.scad
revarbat: However, indentation really *should* be 4-spaces per level.

I used to use a tab for each indentation level and used an editor that showed tabs indented equivalent to 4 spaces. @adrianVmariano used 3 or 4 spaces per indent, depending on the bit of code. This got rather chaotic and problematic eventually. We agreed to standardize on 4-space indents as a result. I converted all the tabs to 4-spaces, and we've been slowly fixing inconsistent space indention as we edit since then.

quaternions.scad
```diff
@@ -214,65 +317,82 @@ function Q_Dist(q1, q2) = norm(q2-q1);
...
+      q2   = _dot<0 ? -q2/norm(q2) : q2/norm(q2),
+      dot = abs(_dot)
+    )
+    ! is_num(u) 
+    ? [for (uu=u) Q_Slerp(q1, q2, uu, dot)]
+    : Q_Slerp(q1, q2, u, dot) 
+  : _dot>0.9995
+    ? _Qnorm(q1 + u*(q2-q1))
+    : let( 
+        theta  = u*acos(_dot),
```
revarbat, line 336: Can `_dot` ever be approximately -1 in this code?  I have had multiple issues with `acos(x)` failing because `x` fell outside of the -1 to 1 domain due to infinitesimal FP rounding errors.  So much so that I pretty much always do a `constrain()` on a value passed to `acos()`

quaternions.scad
```diff
@@ -327,4 +447,216 @@ function Qrot(q,p) =
 module Qrot_copies(quats) for (q=quats) Qrot(q) children();
 
 
+// Function: Q_Rotation()
```
revarbat, line 450: Nice!

quaternions.scad
```diff
@@ -327,4 +447,216 @@ function Qrot(q,p) =
...
+    _Qnorm( _Qset([ R[1][2]-R[2][1], R[2][0]-R[0][2], R[0][1]-R[1][0] ], -r ) )
+  : let( i = max_index([ R[0][0], R[1][1], R[2][2] ]),
+         r = 1 + 2*R[i][i] -R[0][0] -R[1][1] -R[2][2] )
+    i==0 ? _Qnorm( _Qset( [ 4*r, (R[1][0]+R[0][1]), (R[0][2]+R[2][0]) ], (R[2][1]-R[1][2])) ):
+    i==1 ? _Qnorm( _Qset( [ (R[1][0]+R[0][1]), 4*r, (R[2][1]+R[1][2]) ], (R[0][2]-R[2][0])) )
+         : _Qnorm( _Qset( [ (R[2][0]+R[0][2]), (R[1][2]+R[2][1]), 4*r ], (R[1][0]-R[0][1])) );
+
+
+// Function&Module: Q_Rotation_path(q1, n, [q2])
+// Usage as a function:
```
revarbat, line 474: Usage line syntax, to be properly parsed by the documents generator, should have the `:` at the end of the `Usage:` word.  Everything following that is the usage type title.
```
// Usage: As a module
//   path = Q_Rotation_path(q1, n, q2);

File: quaternions.scad [hunk 7]
```diff
@@ -327,4 +447,216 @@ function Qrot(q,p) =
...
+//   qn = Q_Squad(q1,q2,q3,q4,u);
+// Description:
+//   Returns a quaternion that is a cubic spherical interpolation of the quaternions  
+//   q1 and q4 taking the other two quaternions, q2 and q3, as parameter of a cubic 
+//   on the sphere similar to the control points of a Bezier curve.
+//   If u is a number, usually between 0 and 1, returns the quaternion that results
+//   from the interpolation. 
+//   If u is a list of numbers, computes the interpolations for each value in the
+//   list and returns the interpolated quaternions in a list.
+//   The input quaternions don't need to be normalized.
```
revarbat, line 584: Bezier-like quaternion interpolation?  Mind blown!

quaternions.scad
```diff
@@ -214,65 +317,82 @@ function Q_Dist(q1, q2) = norm(q2-q1);
...
-function Q_Angle(q) = 2 * acos(q[3]);
+//   If only q1 is given, returns the angle of rotation (in degrees) of that quaternion.
+//   If both q1 and q2 are given, returns the angle (in degrees) between them.
+//   The input quaternions don't need to be normalized.
+function Q_Angle(q1,q2) =
+  assert(Q_is_quat(q1) && (is_undef(q2) || Q_is_quat(q2)), "Invalid quaternion(s)" )
+  let( n1 = is_undef(q2)? norm(_Qvec(q1)): norm(q1) )
+  is_undef(q2)
+  ? 2 * atan2(n1,_Qreal(q1))
+  : 2 * acos(q1*q2/n1/norm(q2)) ;
```
revarbat, line 395: Good extension of functionality.
The `acos()` really should have it's input `constrain()`ed to -1 to 1, though, because FP rounding errors WILL cause problems otherwise.

RonaldoCMP:
I use some of the remarked formatings. Recently I adopted the formatting of
starting lines with ? and : after recognizing that it makes my codes more
readable especially when the expressions are longer. Maybe I get so used to
it that I use it even for shorter expressions. Anyway, as a general rule I
try to keep the lines shorter so the logic is clearer and  easy grasped. I
can adapt myself to the current BOSL2 style although I think the formats d
and e are much harder to read.

Em sáb., 18 de jul. de 2020 às 03:47, Revar Desmera <
notifications@github.com> escreveu:

RonaldoCMP:
>
>  //   a = Number of degrees to rotate around the axis counter-clockwise, when facing the origin.
> -function QuatX(a=0) = Quat([1,0,0],a);
> +function QuatX(a=0) =
> +  assert( is_num(0*a), "Invalid angle" )
>
> Is the is_num(0*a) instead of the equivalent is_finite(a) for speed?
>
Not exactly. It was not clear to me that is_finite had been defined in the
library. It is clearer to use it although in some places in this file I
intentionally repeat short expressions defined by other file functions to
avoid extra unnecessary checks.

quaternions.scad
RonaldoCMP: It would be preferable but in order to avoid the complement operator (!), something like `non_zero(x)` should be defined as well. Maybe `is_zero` could be defined not only for checking scalars but also of more complex lists. (Note that `approx([0,0],0)` is false.) Something like this:

```
function is_zero(a,eps=EPSILON) =
  (is_num(a) && abs(a)<eps) ||
  (is_list(a) && len(a)>0 && ([]==[for (i=idx(a)) if(! is_zero(a[i] , eps)) 1]));
```
There is room for different criteria but this definition will return true for all the following calls:

```
echo(is_zero([0,0]));
echo(is_zero([0,[0]]));
echo(is_zero([1e-12,0]));
echo(is_zero([0,[0,[1e-12]]]));
echo(! is_zero(1));
echo(! is_zero([1,2,3]));
echo(! is_zero([]));
echo(! is_zero([0,[]]));
echo(! is_zero([1,2,["a"]]));
echo(! is_zero("a"));
echo(! is_zero([0,"a"]));
echo(! is_zero([0,1]));
echo(! is_zero([undef,1]));
echo(! is_zero([1,2,undef]));
echo(! is_zero([0,[0,[undef]]]));
echo(! is_zero([0,[0,[]]]));
```

That is: it is true if and only if all entries of the checking value are numerically near zero. Void list are not zero.

quaternions.scad
RonaldoCMP: Yes, it could but `dot`, the value it has in the recursion, is greater than 0; so the test `_dot>0.9995` will act as a constrain.

quaternions.scad
RonaldoCMP: This function task could be done by Q_Slerp and Q_Matrix4 but it would be very inefficient.

quaternions.scad
RonaldoCMP: Noted.

quaternions.scad
RonaldoCMP: Yes, it is. I can't see myself using this function but for sake of completeness I include it. The same with Q_exp and Q_ln although those are useful to compare with alternative expressions. For instance, quaternion power is usually defined based in exp and ln but I was successful in express a faster alternative without them. 

Do you believe there are things like Squad based on B-splines?

quaternions.scad
RonaldoCMP: You are right, here `constrain()` should be used.

RonaldoCMP:
I will do changes in quaternions.scad agreeing with this discussion. Anything else to add or change?

quaternions.scad
revarbat: Okay, I'm satisfied that this is correct.

quaternions.scad
revarbat: I once proposed support for a full NURBS extension on Beziers, but we decided that would make for a rather complicated interface for minimal extra functionality.

revarbat:
If you want to implement `is_zero()` and/or `nonzero()` in common.scad, or enhance `is_vector()` to support a `nonzero=true` arg in vectors.scad, then go for it.  Otherwise I'll end up doing that after merging your completed PR.

RonaldoCMP:
> If you want to implement `is_zero()` and/or `nonzero()` in common.scad, or enhance `is_vector()` to support a `nonzero=true` arg in vectors.scad, then go for it. Otherwise I'll end up doing that after merging your completed PR.

@adrianVmariano is right, the recursion in `is_zero `is expensive for vectors. A `nonzero `parameter in `is_vector `would be nice for input data checks. For scalar tests, we still need to use `approx`. When we already know that the input `v` is a vector, calling `is_vector(v,nonzero=true)` will require a negligible extra work. I will update `common.scad`.

Regarding quaternions.scad, will you change it according to our discussion here? I had started to do it.

The `acos` expression in Q_angle would be better changed to an `atan2` expression:

```
    is_undef(q2) ? 2 * atan2(n1,_Qreal(q1)) 
    let( q1 = q1/norm(q1),
         q2 = q2/norm(q2) )
    2 * atan2(norm(q1 - q2), norm(q1 + q2));
```

which exempts it from calling `constrain`.

adrianVmariano:
What about a `positive=true` option to is_vector?   I think this is a test I often need.   Also would it make sense to have dual purpose parameters?  So `zero=true` requires that all the entries are zero, `zero=false` requires that all entries are nonzero.  And `zero=undef` imposes no constraint?

revarbat:
You need to make the quaternions.scad changes, commit and push them, then I can approve and merge.

revarbat:
The `atan2()`  change is a good idea.  I kind of like @adrianVmariano's `zero=true/false/undef` idea.  Would `positive=false` mean `<=0`, and `positive=true` mean `>0`?  Would

revarbat:
The `atan2()`  change is a good idea.

I kind of like @adrianVmariano's `zero=true/false/undef` idea.  Would `positive=false` mean `<=0`, and `positive=true` mean `>0`?  That would imply the need for `negative=true/false/undef` as well.

adrianVmariano:
Yes, I think that makes sense, that `positive=false` means all entries `<=0` and then we have `negative` as well, which would avoid the need for the more clumsy `nonnegative`.  

So would the tests for zero use eps but the tests for positive/negative us comparisons to exact zero?   

Do we need to add this check to anything other than vectors?   Is there (despite my doubt) utility in writing the slower recursive general checker, `check_all(list, poistive=true)` that would follow the same syntax?

RonaldoCMP:
I usually prefer many specific functions than a few with several options that I have to remember the parameter names and its effects. Remember that we already have the option of specifying (or not) the length of a vector, which is invaluable. A nonzero option would be a nice and useful addition. Anything else should have other functions to check. For instance, a function vector_sign(v) returning a string or scalar for positive, negative, non negative, none, etc.

adrianVmariano:
I think that the issue of having to remember parameter names is not really a valid critique of the proposed approach.  You always have to remember stuff.  It's just a question of what stuff you have to remember.  If you have a dozen functions for checking vector characteristics then you have to remember the names of those functions.  (And that's actually harder to look up in the manual if you can't remember them.)   In the proposal of a vector_sign function you have to remember the name of the function and what the possible return values are to make the comparison.  (That is, instead of `is_vector(v,positive=true)` it's `is_vector(v) && vector_sign(v)=='positive'`.  Seems to me the memory burden is higher in this second case, actually.  

It seems to me that the real question is making a design choice and then trying to make things generally consistent.   I think that from a design perspective, having a `vector_sign` function that returns a string is less elegant and less consistent in general with how we have done things.   It's also more difficult to code efficiently, I think.  And actually now that I think about it's actually kind of troublesome to use.  If we have `vector_sign` and it returns one of ['positive','nonnegative','zero','negative','nonpositive'] then I have a problem because positive is also nonnegative.  So do I return a list of properties, like `['positive','nonnegative']`?  Or do I return the strictest one, just 'postitive'?   Then you have to test for postitive and nonnegative to catch all the nonnegative cases, so that doesn't make sense.  So the only way it makes sense it to return a property list.  This is ineffficient because it means I have to compute properties I didn't care about, and then I have to use `in_list` to check the result.  And `in_list('nonnegative',vector_sign(v))` to me seems to be a lot more memory burden to remember than other ideas.  It's also harder to read.  

I see the following reasonable paths:  We add this functionality to `is_vector` as Revar suggested.  We could add similar functionality to other tests as well.   Instead of (or in addition) we write a single point checker, so not `vector_sign` but `check_sign` or something that takes a similar set of args, so `check_sign(v,positive=true)`, or we write a collection of checker functions, so `is_positive(v)`, `is_negative(v)`, `is_nonnegative(v)`, etc.

revarbat:
I kind of like the idea of `is_vector()` having `positive=` or `negative=`, but in truth, the deciding issue is if we can still have a relatively fast check with those features.  If not, then the `check_sign()` idea seems indicated.

adrianVmariano:
Isn't the overhead for adding `positive=` and `negative=` simply the time to do two checks for undef?  The condition would be `<current is_vector stuff> && (is_undef(positive) || ....) && (is_undef(negative) || ....)`

revarbat:
Latest commit is looking good, @RonaldoCMP.  Is this the final code you want to commit?

revarbat:
Actually, what exactly does it mean for a vector to be positive?  That all components are positive?

revarbat:
I think the zero/nonzero/positive/negative/etc checks ideas should be moved to it's own issue at this point.

RonaldoCMP:
> Latest commit is looking good, @RonaldoCMP. Is this the final code you want to commit?

It is fine for me but I was not able Pull a Request as I did last time.

quaternions.scad
revarbat: Deferring this into it's own issue.

revarbat:
Your commits were automatically added to this PR, since you added them to the same git branch (master).  Once I merge the PR, you will be able to make another PR from your master branch.  And I'm merging now.

PR #213, opened 2020-07-24
Reviews of input data check and format

RonaldoCMP:
Here is an extensive review of common.scad, vectors.scad, arrays.scad and their regression test codes. The changes were mostly in input data check and format. Some few functions got more changes either to adapt them to the new `asserts `or because a better (faster, simpler) code was found. Some inserted `asserts `have side effects on other file tests:

1. test_math.scad fails trying to do a `vmul` of vectors with different lengths;
2. the same happens with test_shapes.scad and test_shapes2d.scad;

I believe that trying to multiply vectors of different length should be considered an error. I didn't dig into the offending codes yet to see what is wrong.

Another change was moving `add_scalar` from vectors.scad to arrays.scad. It is in fact an array operation. Some few documentation examples and test codes were added. Some function argument names were changed to keep them consistent with the names they have in other similar functions. That name changes were not as extensive as I would like but are advisable to favor users.

Most, but not all, changed function codes are followed by a comment with a short summary of the changes done.  Those comments always start with:
//***

I can eliminate those comments before a merge is done.

### DIFF
--- a/arrays.scad
+++ b/arrays.scad
@@ -795,11 +795,6 @@ function _sort_vectors4(arr) =
 		concat( _sort_vectors4(lesser), equal, _sort_vectors4(greater) );
 		
 
-function _sort_eval(array, eval) =
-    let( ival  = [for(i=[0:len(array)-1]) [eval[i], i] ],
-         sival = _sort_vectors1(arr) )
-		[for(yi=sival) array[yi[1]] ];
-
 function _sort_general(arr, idx=undef) =
     (len(arr)<=1) ? arr :
     let(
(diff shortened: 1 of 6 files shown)


RonaldoCMP:
I did some correction in the code but am unable to submit it again. How I "delete" an unsuccessful PR?

revarbat:
Lemme try deleting it.

revarbat:
I can close it without merging.

revarbat:
You probably need to do a fit pull from the upstream master branch.

PR #219, opened 2020-07-28
hull fix and regression tests

adrianVmariano:


revarbat:
I'll fix the regression test.  I think it looks valid.

PR #234, opened 2020-08-16
Geometry full review and minor edits

RonaldoCMP:


RonaldoCMP:
Yes, the last definition is the one to retain. I have been doing this to
keep the old version for your reference and quicker evaluation. If nobody
complains I strip the old one in a second round.

Em ter., 18 de ago. de 2020 às 21:48, Revar Desmera <
notifications@github.com> escreveu:

revarbat:
Ahh.  Okay, it's actually easier for me to read the diffs, so the original definitions aren't needed.

RonaldoCMP:
I usually also write some comments along the file with questions about the code. If nobody replies my comments I strip them off in the second turn.

I have already edited the file excluding the repeated definitions. May I request a new pull?

RonaldoCMP:
I guess I have removed all the duplicate definitions and besides solved one conflict in rounding.

geometry.scad
```diff
@@ -1606,11 +1734,29 @@ function centroid(poly) =
...
     // Does the point lie on any edges?  If so return 0.
-    sum([for(i=[0:1:len(path)-1]) let(seg=select(path,i,i+1)) if(!approx(seg[0],seg[1],eps=eps)) point_on_segment2d(point, seg, eps=eps)?1:0]) > 0? 0 :
+    let(
+        on_brd = [for(i=[0:1:len(path)-1]) 
+                    let( seg = select(path,i,i+1) ) 
+                    if( !approx(seg[0],seg[1],eps=eps) ) 
+                        point_on_segment2d(point, seg, eps=eps)? 1:0 ]
+        )
+    sum(on_brd) > 0? 0 :
     // Otherwise compute winding number and return 1 for interior, -1 for exterior
```
revarbat, line 1748: Do we need to have an option to specify whether we want to use Even-Odd or Non-Zero winding rules?

geometry.scad
```diff
@@ -1772,13 +1915,27 @@ function _split_polygon_at_z(poly, z) =
...
     _i>=len(xs)? polys :
     split_polygons_at_each_x(
         [
             for (poly = polys)
             each _split_polygon_at_x(poly, xs[_i])
         ], xs, _i=_i+1
     );
+    
+//***
+// all the functions split_polygons_at_ may generate non simple polygons even from simple polygon inputs:
```
revarbat, line 1930: Please file an issue for this.

tests/test_geometry.scad
```diff
@@ -29,42 +167,28 @@ module test_point_on_segment2d() {
     assert(point_on_segment2d([ 10, 10], [[-10,-10], [10,10]]) == true);
     assert(point_on_segment2d([ 15, 15], [[-10,-10], [10,10]]) == false);
 }
-test_point_on_segment2d();
+*test_point_on_segment2d();
```
revarbat, line 170: Does the module actually execute, but have the geometry nullified by `*`, or does it not run at all?

common.scad
```diff
@@ -289,13 +289,28 @@ function get_anchor(anchor,center,uncentered=BOT,dflt=CENTER) =
...
 //   dflt = Value to return if all other values given are `undef`.
 function get_radius(r1=undef, r2=undef, r=undef, d1=undef, d2=undef, d=undef, dflt=undef) = (
-    !is_undef(r1)? assert(is_undef(r2)&&is_undef(d1)&&is_undef(d2), "Conflicting or redundant radius/diameter arguments given.") r1 :
-    !is_undef(r2)? assert(is_undef(d1)&&is_undef(d2), "Conflicting or redundant radius/diameter arguments given.") r2 :
-    !is_undef(d1)? d1/2 :
-    !is_undef(d2)? d2/2 :
-    !is_undef(r)? assert(is_undef(d), "Conflicting or redundant radius/diameter arguments given.") r :
-    !is_undef(d)? d/2 :
-    dflt
+    !is_undef(r1)
```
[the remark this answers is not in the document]
RonaldoCMP: Good idea. Yet, `num_defined()` doesn't call `sum()` but it has 2 main issues:
1. it would break if parameter `v` isn't a list;
2. it would break if an user call it with more than 1 parameter (this is common to many other recursive functions).

The former is solved by an `assert`.  To solve the later we have 3 alternatives:
1. call an internal recursive function to do the job;
2. change it to a C-like for avoiding recursion;
3. call `sum()` which is fast and non-recursive for vectors.

I prefer the later.

geometry.scad
```diff
@@ -21,85 +21,72 @@
...
+         de = edge[1]-edge[0],
+         ne = norm(de) ) 
+    ( dp*de >= -eps*ne ) 
+    && ( (dp-de)*de <= eps*ne )              // point projects on the segment
+    && _dist(point-edge[0],unit(de))<eps;   // point is on the line 
+    
+    
+//Internal - distance from point `d` to the line passing through the origin with unit direction n
+//_dist works for any dimension
+function _dist(d,n) = norm(d-(d * n) * n);
```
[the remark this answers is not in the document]
RonaldoCMP: Noted. May be `_dist2line()`

geometry.scad
```diff
@@ -824,25 +879,29 @@ function plane_from_points(points, fast=false, eps=EPSILON) =
...
-            if (!coplanar(plane,pt,eps=eps)) 1
-        ] == []
-    ) all_coplanar? plane :
-    undef;
+    ) 
+    fast? plane: coplanar(poly,eps=eps)? plane: [];
 
+//***
+// I don't see why this function uses a criterium different from plane_from_points.
+// In practical terms, what is the difference of finding a plane from points and from polygon?
```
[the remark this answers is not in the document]
RonaldoCMP: Understood. Should I drop it?

geometry.scad
```diff
@@ -1156,37 +1254,45 @@ function in_front_of_plane(plane, point) =
...
+            dang1 = vector_angle(tp1-cp,pt2-cp),
+            dang2 = vector_angle(tp2-cp,pt2-cp)
+            ) 
+        [cp, n, tp1, tp2, dang1, dang2];
 
 
 // Function: find_circle_3points()
 // Usage:
-//   find_circle_3points(pt1, pt2, pt3);
+//   find_circle_3points(pt1, [pt2, pt3]);
```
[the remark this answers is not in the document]
RonaldoCMP: Noted. Perhaps we could use:

// Usage:
//   find_circle_3points(pt1, <pt2, pt3>);

to refer to optional parameters. It is less conventional though.

geometry.scad
RonaldoCMP: Done. The case I reported as a fail in `split_polygons_at_  ` functions was wrong because it inputs a polygon and not a list of polygons  as required.

tests/test_geometry.scad
RonaldoCMP: Yes. The * prevents the module run at all. Instead of reordering the modules, I have preferred to do the ordered calls at the beginning. The * are there for the case someone prefers to change it to the previous schema.

RonaldoCMP:
There is a question I included in the code of `geometry.scad` you haven't mentioned: why sort the indices in `plane_from_points `and other similar functions?

geometry.scad
```diff
@@ -875,7 +936,34 @@ function plane_transform(plane) =
     let(
         n = plane_normal(plane),
         cp = n * plane[3]
-    ) rot(from=n, to=UP) * move(-cp);
+        ) 
+    rot(from=n, to=UP) * move(-cp);
+
+
+// Function: projection_on_plane()
```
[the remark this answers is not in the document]
RonaldoCMP: Well, they have different purposes: `project_plane ` generates 2D points in a local coordinate system on the plane while `projection_on_plane`  generates the list of the projected points in the same 3D coordinate system. The similarity of their names may be misleading though.

geometry.scad
RonaldoCMP: I am not sure I understand the options.

common.scad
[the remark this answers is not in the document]
revarbat: Good idea.

geometry.scad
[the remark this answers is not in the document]
revarbat: Sounds reasonable.

geometry.scad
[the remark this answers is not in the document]
revarbat: Unless `plane_from_polygon()` is particularly slower, I think we should just go with that one.

geometry.scad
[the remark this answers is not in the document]
revarbat: Oh!  I missed that.  Hmm.  `snap_to_plane()`?

geometry.scad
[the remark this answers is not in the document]
revarbat: That looks clean.

geometry.scad
revarbat: Even-Odd Winding Rule for insideness: https://en.wikipedia.org/wiki/Even–odd_rule
Nonzero Winding Rule: https://en.wikipedia.org/wiki/Nonzero-rule

revarbat:
I honestly don't recall why the indices are sorted.

geometry.scad
revarbat: 440px-Even-odd_and_non-zero_winding_fill_rules
Even-Odd on the bottom left.  Nonzero on the bottom right.

RonaldoCMP: Aha, I knew the rules, not the names. It seems that OpenSCAD uses the Non-zero rule in 2D. I will think how to implement both.

RonaldoCMP:
> I honestly don't recall why the indices are sorted.

I think I may found a reason: it is a way to standardize the output plane. If `P` is a plane, `-P` represents the same plane. The regression test will fail if the computed plane representation is not equal to the right hand side of the test. This happens with quaternions too. My work around to that was to compare standard representations:

`assert( standardize(<computed quaternion>)== standardize(<expected quaternion>);`

 The standardization function I defined to quaternions may be used here with planes:
```
function standardize(v) =  v==[]? [] : sign([for(vi=v) if( ! approx(vi,0)) vi, 0 ][0])*v;

standardize([0,-1,2,3]) == [0,1,-2,-3]
standardize([0,1,-2,-3]) == [0,1,-2,-3]
```
I think the regression test is the place to do the standardization otherwise it will be an additional operation on each function outputting a plane or quaternion only to simplify the tests.

geometry.scad
RonaldoCMP: I have an implementation but to insert it now would require changes in calls from other files. It will be left for a second round.

revarbat: I'm fine with deferring this.

revarbat:
Hmm.  Noticing that commit ee80b1d08f2e26b2350e2cf9512e4729ad680a73 (Merge remote-tracking branch 'upstream/revarbat_dev') looks like it reverts @adrianVmariano's recent rounding.scad changes.

revarbat:
Docs generation check fails.

Basically `bent_cutout_mask()` example 11 has a syntax error, I think.

```
ERROR: Parser error in file "/home/runner/work/BOSL2/BOSL2/rounding.scad", line 679: syntax error

Can't parse file 'tmp_bent_cutout_mask_11_png.scad'!

////////////////////////////////////////////////////
// rounding.scad: tmp_bent_cutout_mask_11_png.scad for images/rounding/tmp_bent_cutout_mask_11.png
////////////////////////////////////////////////////
include <BOSL2/std.scad>
include <BOSL2/debug.scad>
include <BOSL2/rounding.scad>
r=25;
rot(-90) {
  $fn=128;
  difference(){
    tube(or=r, wall=2, h=45);
    bent_cutout_mask(r-1, 2.1, apply(back(15),subdivide_path(round_corners(star(n=7,ir=5,or=10), cut=flatten(repeat([0.5,0],7))),14*15,closed=true)));
  }
}
////////////////////////////////////////////////////
```

PR #239, opened 2020-08-18
Fixed VNF attachment normal vector.

revarbat:


RonaldoCMP:
plane_normal() is an unit vector.

PR #242, opened 2020-08-22
Fix diff/intersect when some tags are hidden

kelvie:
When the first operand is hidden, some strange things happen, so this guards
against that.

Example:

    include <BOSL2/std.scad>

    part_to_show = "A"; // [A,B,C]
    part_to_hide = "B"; // [A,B,C]

    module my_part() {
        tags("A") left(10) cuboid(10);
        tags("B") diff("neg", "B") right(10) {
            cuboid(10);
            cyl(d=10, h=10, $tags="neg");
        }
        tags("C") fwd(10) cuboid(10);
    }

    show(part_to_show) hide(part_to_hide) my_part();

Tag `B` here acts very strangely when it is supposed to be hidden, and never
hides completely.

### DIFF
--- a/attachments.scad
+++ b/attachments.scad
@@ -1235,17 +1235,20 @@ module show(tags="")
 //   }
 module diff(neg, pos=undef, keep=undef)
 {
-    difference() {
-        if (pos != undef) {
-            show(pos) children();
-        } else {
-            if (keep == undef) {
-                hide(neg) children();
+    // Don't perform the operation if the current tags are hidden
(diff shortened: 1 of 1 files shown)


revarbat:
As best as I can tell, the issue is that the `diff()` call internals are doing a `show("B") children();` inside its `difference()`, which is overriding your `hide()`.

It's looks to me that the current behavior of nested `show()` and `hide()` may be flawed.  I'll take a look at your proposed patch later, when I don't have a massive headache.

PR #246, opened 2020-08-27
Added is_zero(), is_positive(), is_negative, is_nonpositive(), is_nonnegative().

revarbat:


RonaldoCMP:
Should not the last test in `is_zero()` be `is_num(x) ? approx(x,0,eps) : false` ? 
Anyway a call of `approx()` in `is_zero() `adds 5 more tests whose outcome are know at that point. Why not just

```
function is_zero(x, eps=EPSILON) =
    is_list(x)? (x != [] && [for (xx=x) if(!is_zero(xx,eps=eps)) 1] == []) :
    is_num(x)? abs(x)<eps :
    false;
```
? 
Am I missing something?

PR #259, opened 2020-09-02
Fixes #258 - [walls] - Implement shell module

### LINKED ISSUE #258: Module for generating shell 3D models
**Is your feature request related to a problem? Please describe.**
A common pattern in 3D models consists of a base 3D shape with vertical shells (walls) like if base was extruded but hollow in the inside . It seems there is no way to do so out of the box.

**Describe the solution you'd like**
An operator that receives as input the base 3D model and outputs the same object augmented with shell walls.

**Describe alternatives you've considered**
Doing this by hand might be a quite annoying repetitive task.

**Example Code**
Usage should be like this

```
shell(<width>, <height>) <3D mnodel>
```

For instance

```
shell(0.3, 10)
    hull() {
           translate(3,3,0)
               sphere(r=3);
           translate(7,3,0)
               sphere(r=3);
           translate(3,12,0)
               sphere(r=3);
           translate(7,12,0)
               sphere(r=3);
    }
```

... should render ...

rounded shell cube base

hollow body with shells

A similar example for chamfer

```
module pyramid(height) {
    polyhedron(points=[[0, 0, height], [height, 0, 0], [0, -height, 0], [-height, 0, 0], [0, height, 0]], faces=[[0, 1, 2], [0, 2, 3], [0, 3, 4], [0, 4, 1], [1,2,3,4]]);
}

shell(2, 20)
    hull() {
        translate([-20, -20])
            mirror([0,0,1]) pyramid(4);
        translate([20, -20])
            mirror([0,0,1]) pyramid(4);
        translate([-20, 20])
            mirror([0,0,1]) pyramid(4);
        translate([20, 20])
            mirror([0,0,1]) pyramid(4);
    }
```

image

image

**Additional context**
The example explained above arose while discussing a real-life application for OTTO-DIY [body](https://wikifactory.com/+OttoDIY/otto-diy/file/3Dprint/OttoDIY_body_V10.stl) and [head](https://wikifactory.com/+OttoDIY/otto-diy/file/3Dprint/OttoDIY_head_V10.stl).

olemis:
Fixes #258 

**Changes:**
  - Add `shell` module in `walls.scad`

**Context**
Operator that receives as input the base 3D model and outputs the same object augmented with shell walls.

### DIFF
--- a/mutators.scad
+++ b/mutators.scad
@@ -603,5 +603,33 @@ module rainbow(list, stride=1)
     }
 }
 
+// Module: shell()
+//
+// Description:
+//   Operator that receives as input the base 3D model and outputs the same object augmented with shell walls.
+//
+// Usage:
+//   shell(w, l) base;
+//
+// Arguments:
+//   w = Width (thickness) of the shell.
+//   l = Length of the shell.
+//   base = 3D model representing object base
+//
+// Example: Square box
+//  shell(1, 20) cube([20, 20, 1]);
+module shell(width, height) {
+    union() {
+        linear_extrude(height)
+            difference() {
+                projection()
+                    children(0);
+                offset(-width)
+                    projection() children(0);
+            }
+        children(0);
+    }
+}
+
 
 // vim: expandtab tabstop=4 shiftwidth=4 softtabstop=4 nowrap
--- a/walls.scad
+++ b/walls.scad
@@ -529,6 +529,9 @@ module corrugated_wall(h=50, l=100, thick=5, strut=5, wall=2, anchor=CENTER, spi
 //   w = Width (thickness) of the shell.
 //   l = Length of the shell.
 //   base = 3D model representing object base
+//
+// Example: Square box
+//  shell(1, 20) cube([20, 20, 1]);
 module shell(width, height) {
     union() {
         linear_extrude(height)
--- a/walls.scad
+++ b/walls.scad
@@ -517,6 +517,30 @@ module corrugated_wall(h=50, l=100, thick=5, strut=5, wall=2, anchor=CENTER, spi
     }
 }
 
+// Module: shell()
+//
+// Description:
+//   Operator that receives as input the base 3D model and outputs the same object augmented with shell walls.
+//
+// Usage:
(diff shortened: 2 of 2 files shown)


olemis:
@revarbat what would be the right way to add tests and docs ?

revarbat:
Okay, so a few issues:
- The name `shell()` is maybe unclear that you're making an open-top container.
- It makes more sense to pass a 2D shape to it, than to have it make projections of a 3D shape.
- If you DO intend to have it take a 3D shape, I would expect it to do a synthetic difference Minkowski, instead of using `offset()` on 2D `projection()`s of the 3D shape.
- Why `children(0)` (the first child) instead of `children()` (all children).

revarbat:
If you want to make it just take a 2D shape, I propose this (though the name needs work still):
```openscad
module shell(h, wall=2, convexity=4) {
    linear_extrude(height=wall, convexity=convexity, center=false) children();
    linear_extrude(height=h+wall, convexity=convexity, center=false) {
        difference() {
            children();
            offset(r=-wall) children();
        }
    }
}
```

adrianVmariano:
I'm not convinced that this is a sufficiently powerful operation to be worth adding.  It also does not do the thing that was described in the issue it claims to solve:  in particular, you don't get chamfers or roundings on the box as shown in the pictures.    You need to use offset_sweep to make such a box for the general case, and I thought about making a box() module based on offset_sweep, but it seemed like there were two reasons not to do this:  (1) it's very easy to make the desired box with two calls to offset_sweep().   (2) the box() module that would be required would have very complicated arguments...and basically just all the arguments needed for two calls to offset_sweep, so it doesn't provide a significant advantage.

olemis:
> it claims to solve: in particular, you don't get chamfers or roundings on the box as shown in the pictures.

There is a misunderstanding. The round corners and chamfer are not features inherent to proposed `shell` module. They are part of `children()` geometry. In practice they are just examples we rely upon to explain operators to teenager students of a course we offer.

revarbat:
This isn't going to be implemented quite like this.

PR #265, opened 2020-09-08
Added all_nonzero().  Added is_vector(all_nonzero=).

revarbat:


RonaldoCMP:
I am not sure I agree with this change in the definition of `is_vector()`. That test is broadly used and the last two definition changes impose two more tests in it: for `zero` and for `all_nonzero`. Those two parameters will be used in rare cases that could be checked with other functions. For instance, the following expressions are all true:

```
is_vector(v, zero=true)  ==  ( is_vector(v) && norm(v)<eps )
is_vector(v,zero=false) == ( is_vector(v) && norm(v)>=eps )
is_vector(v,all_nonzero=true) == ( is_vector(v) && all_nonzero(v) )
```
To me, the right hand expressions above are clearer and succinct enough to be preferred to the left hand side expressions. Many test functions like `is_vector()` will have to include additional tests to avoid future warnings trying to capture inappropriate expressions with `undef`. For better performance, we should keep the tests to a minimum for the general more often cases.

PR #266, opened 2020-09-09
Solving bugs in plane operations; extending tests in test_geometry

RonaldoCMP:
Correct bugs in plane operations for non-normalized planes.
Includes in test_geometry all missing tests for functions dealing with plane.
Correct small bug in regions.scad.

### DIFF
--- a/geometry.scad
+++ b/geometry.scad
@@ -924,7 +924,7 @@ function plane_offset(plane) =
 //   stroke(xypath,closed=true);
 function plane_transform(plane) =
     let(
-		    plane = normalize_plane(plane),
+        plane = normalize_plane(plane),
         n = point3d(plane),
         cp = n * plane[3]
         ) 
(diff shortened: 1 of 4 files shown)


geometry.scad
```diff
@@ -1008,19 +1011,23 @@ function closest_point_on_plane(plane, point) =
...
+    )
+    approx(b,0,eps)                          // is  (line[1]-line[0]) "parallel" to the plane ?
+    ? approx(a,0,eps)                        // is line[0] on the plane ?
+       ? [line,undef]                        // line is on the plane
+       : undef                               // line is parallel but not on the plane
+    : [ line[0]-a/b*(line[1]-line[0]), -a/b ];
+    
+    
+// Function: normalize_plane()
+// Usage: normalize_plane(plane)
```
revarbat, line 1025: The function usage description (`normalize_plane(plane)`) needs to be on its own line. The text following on the same line as the `Usage:` header is an optional usage type title like "When called as a function".

```openscad
// Usage:
//   p = normalize_plane(plane);
```

revarbat (changes requested):

RonaldoCMP:
Done.

PR #279, opened 2020-09-27
Change line_of to work with n == 1

kelvie:
The use case for this is something like this, to ensure there is always a screw
hole every 40mm:

    n = ceil(width/40);
    xcopies(n=n, l=(1 - 1/n)*width)
        screw_hole();


revarbat:
Thanks for the find!  There's a cleaner one-line fix for this, though.  I've already patched this, and improved the docs and argument checking while I'm at it.

PR #322, opened 2020-12-01
fixed modulated_circle() to work as advertised

plut:
The doc for this module suggested a list of pairs, the default argument was a list of 1 number, and what would actually work was list of triples.

This patch fixes the module to accept list of either pairs or triples, updates the doc accordingly, and gives an accepted default argument.

### DIFF
--- a/paths.scad
+++ b/paths.scad
@@ -747,16 +747,20 @@ function assemble_path_fragments(fragments, eps=EPSILON, _finished=[]) =
 // Arguments:
 //   r = Radius of the base circle. Default: 40
 //   d = Diameter of the base circle.
-//   sines = array of [amplitude, frequency] pairs, where the frequency is the number of times the cycle repeats around the circle.
+//   sines = array of [amplitude, frequency] pairs or [amplitude, frequency, phase] triples, where the frequency is the number of times the cycle repeats around the circle.
 // Example(2D):
 //   modulated_circle(r=40, sines=[[3, 11], [1, 31]], $fn=6);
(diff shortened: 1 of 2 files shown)


paths.scad
```diff
@@ -747,16 +747,20 @@ function assemble_path_fragments(fragments, eps=EPSILON, _finished=[]) =
...
 //   modulated_circle(r=40, sines=[[3, 11], [1, 31]], $fn=6);
-module modulated_circle(r, sines=[10], d)
+module modulated_circle(r, sines=[[1,1]], d)
 {
     r = get_radius(r=r, d=d, dflt=40);
-    freqs = len(sines)>0? [for (i=sines) i[1]] : [5];
+    assert(is_list(sines)
+        && all([for(s=sines) is_vector(s,2) || is_vector(s,3)]),
+        "sines must be given as a list of pairs or triples");
+    sines_ = [for(s=sines) [s[0], s[1], len(s)==2 ? 0 : s[2]]];
```
revarbat, line 759: This can actually be written as:
```openscad
sines_ = path3d(sines);
```
It'll do the same thing.  However, I can see a reason not to use `path3d()` as it's notionally a different thing, even if it's result is identical.

revarbat: Actually, `path3d()` may test for `is_path()`, which would not like different length [freq, amp, offset] values.  Also, it may require at least two of them.

revarbat:
The regression failure is not this PR's fault, but my fault, which was fixed externally to this PR.

PR #326, opened 2020-12-03
Plut dev

plut:
This extends the `half_of()` (etc.) modules to also work as functions (when applied to paths/regions/VNFs). It contains an implementation of intersection of a VNF with a half-space.

This also contains a `get_named_args()` function that might be useful for defining functions that may be called with named and/or positional arguments.

### DIFF
--- a/common.scad
+++ b/common.scad
@@ -312,38 +312,67 @@ function get_height(h=undef,l=undef,height=undef,dflt=undef) =
     assert(num_defined([h,l,height])<=1,"You must specify only one of `l`, `h`, and `height`")
     first_defined([h,l,height,dflt]);
 
-// Function: get_named_args(positional, named, _undef)
+// Function: get_named_args()
 // Usage:
 //   function f(pos1=_undef, pos2=_undef,...,named1=_undef, named2=_undef, ...) = let(args = get_named_args([pos1, pos2, ...], [[named1, default1], [named2, default2], ...]), named1=args[0], named2=args[1], ...)
 // Description:
 //   Given the values of some positional and named arguments,
-//   returns a list of the values assigned to named arguments,
-//   in the following way:
-//    - All named arguments which were explicitly assigned in the
-//    function call take the value provided.
-//    - All named arguments which were not provided by the user are
-//    affected from positional arguments; the priority order in which
-//    these are assigned is given by the `priority` argument, while the
-//    positional assignation is done in the order of the named arguments.
-//    - Any remaining named arguments take the provided default values.
-//  If only k positional arguments are used, then the k named values
-//  with lowest 'priority' value (among the unassigned ones) will get them.
-//  The arguments will be assigned in the order of the named values.
-//  By default these two orders coincide.
+//   returns a list of the values assigned to named parameters.
+//   in the following steps:
+//   - First, all named parameters which were explicitly assigned in the
+//      function call take their provided value.
+//   - Then, any positional arguments are assigned to remaining unassigned
+//     parameters; this is governed both by the `priority` entries
+//     (if there are `N` positional arguments, then the `N` parameters with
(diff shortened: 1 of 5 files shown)


mutators.scad
```diff
@@ -64,13 +64,20 @@ module bounding_box(excess=0) {
...
 
 
-// Module: half_of()
+// Function&Module: half_of()
 //
-// Usage:
+// Usage: as module
 //   half_of(v, [cp], [s]) ...
+// Usage: as function
+//   half_of(v, [cp], p, [s])...
```
revarbat, line 72: Probably should be:
```
// Usage: as function
//   vnf = half_of(v, <cp>, p, <s>);
```

mutators.scad
```diff
@@ -64,13 +64,20 @@ module bounding_box(excess=0) {
 }
 
 
-// Module: half_of()
+// Function&Module: half_of()
 //
-// Usage:
+// Usage: as module
 //   half_of(v, [cp], [s]) ...
```
revarbat, line 70: We used to use [foo] to denote optional args in Usage, but we ran into a lot of cased where we wanted to indicate the passing of a list argument. Thus we've been moving the Usage optional args to using the syntax <foo>.

common.scad
```diff
@@ -312,7 +312,66 @@ function get_height(h=undef,l=undef,height=undef,dflt=undef) =
...
 
-
+// Function: get_named_args(positional, named, _undef)
+// Usage:
+//   function f(pos1=_undef, pos2=_undef,...,named1=_undef, named2=_undef, ...) = let(args = get_named_args([pos1, pos2, ...], [[named1, default1], [named2, default2], ...]), named1=args[0], named2=args[1], ...)
+// Description:
+//   Given the values of some positional and named arguments,
+//   returns a list of the values assigned to named arguments,
+//   in the following way:
+//    - All named arguments which were explicitly assigned in the
```
revarbat, line 322: Markdown wants this indented by one less space, I believe.

common.scad
```diff
@@ -312,7 +312,66 @@ function get_height(h=undef,l=undef,height=undef,dflt=undef) =
...
+//    function call take the value provided.
```
revarbat, line 323: And list item continuations like this indented by three more spaces.

mutators.scad
```diff
@@ -64,13 +64,20 @@ module bounding_box(excess=0) {
...
+// Usage: as function
+//   half_of(v, [cp], p, [s])...
 //
 // Description:
 //   Slices an object at a cut plane, and masks away everything that is on one side.
+//   * Called as a function with a path in the `p` argument, returns the
+//   intersection of path `p` and given half-space.
```
revarbat, line 77: Markdown will want this indented four more spaces as a continuation of the list item.

vnf.scad
```diff
@@ -827,5 +827,130 @@ module vnf_validate(vnf, size=1, show_warns=true, check_isects=false) {
...
+// Section: VNF transformations
+//
+// Function: vnf_halfspace(halfspace, vnf)
+// Usage:
+//   vnf_halfspace([a,b,c,d], vnf)
+// Description:
+//   returns the intersection of the VNF with the given half-space.
+// Arguments:
+//   halfspace = half-space to intersect with, given as the four
+//   coefficients of the affine inequation a*x+b*y+c*z+d ≥ 0.
```
revarbat, line 839: We have a plane definition used in geometry.scad where `[A,B,C,D]` is `Ax+By+Cz=D`.  Can we use this for consistency?  The normal of the plane would specify which side of the plane (or halfspace) to use.

revarbat: Also, all arguments in the Arguments: block have to be one-line-per-argument with the syntax `name = description`

PR #420, opened 2021-02-17
Fix function header typos interfering with docs generation

martincizek:


revarbat:
Good catches. Coincidentally, I'm currently rewriting my docs generator, and the new one just caught these yesterday. I'm not quite ready to commit my new docs generator, so I'll merge this in the meantime.

PR #468, opened 2021-03-13
Create bottlecaps_adapters.scad

westminsterflip:
General purpose bottle cap, neck, and threaded adapters.  Modified from bottlecaps.scad.


revarbat:
The CheckDocs fail above is due to a workflow issue, not a problem with the submission.
That said, this PR is submitting a new file instead of an alteration to the original file.

PR #474, opened 2021-03-15
Fixed worm() face normals.

revarbat:


revarbat:
I'm not actually happy with the jagginess of the worm surface, but the real solution will require a rewrite.  This at least makes for a functional worm.

PR #483, opened 2021-03-25
Minor corrections in geometry.scad

RonaldoCMP:


common.scad
```diff
@@ -294,6 +298,7 @@ function first_defined(v,recursive=false,_i=0) =
 //   val = one_defined(vals, names, <dflt>)
 // Topics: Undef Handling
 // See Also: default(), first_defined(), num_defined(), any_defined(), all_defined()
+//   one_defined(vars, names, <required>)
```
revarbat, line 301: This cannot be part of 'See Also:', which does not allow multiple lines.  Should it be a few lines above, in the 'Usage' block?

revarbat:
I'm resubmitting the resolved merge as a new PR.

PR #487, opened 2021-03-30
slice() and select() refactoring.

revarbat:
Brought slice() in line with select() indexing, without wrapping.
Replaced a lot of select() and slice() calls with last(), list_head(), and list_tail() calls.

### DIFF
--- a/paths.scad
+++ b/paths.scad
@@ -183,7 +183,7 @@ function path_length(path,closed=false) =
 function path_segment_lengths(path, closed=false) =
     [
         for (i=[0:1:len(path)-2]) norm(path[i+1]-path[i]),
-        if (closed) norm(path[0]-select(path,-1))
+        if (closed) norm(path[0]-last(path))
     ]; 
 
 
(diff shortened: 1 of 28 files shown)


vnf.scad
```diff
@@ -481,7 +481,7 @@ function _triangulate_planar_convex_polygons(polys) =
         tris = [for (poly=polys) if (len(poly)==3) poly],
         bigs = [for (poly=polys) if (len(poly)>3) poly],
         newtris = [for (poly=bigs) select(poly,-2,0)],
-        newbigs = [for (poly=bigs) select(poly,0,-2)],
```
adrianVmariano, line 484: Maybe leave this one as it is because it shows more directly how newbigs and newtris relate?

paths.scad
```diff
@@ -872,12 +872,12 @@ function assemble_a_path_from_fragments(fragments, rightmost=true, startfrag=0,
...
+        fragend = last(foundfrag),
         hits = [for (i = idx(path,e=-2)) if(approx(path[i],fragend,eps=eps)) i]
     ) hits? (
         let(
             // Found fragment intersects with initial path
-            hitidx = select(hits,-1),
+            hitidx = last(hits),
             newpath = list_head(path,hitidx),
             newfrags = concat(len(newpath)>1? [newpath] : [], remainder),
             outpath = concat(slice(path,hitidx,-2), foundfrag)
```
adrianVmariano, line 883: slice call needs update?

revarbat: Actually, I changed this one back to `slice()`, just in case `hitidx` is ever that near the end of the list.

PR #506, opened 2021-04-17
Cleanup for bottlecaps_adapters.scad

revarbat:


westminsterflip:
Why were .1 differences/additions added in the difference in thread_adapter_NC and thread_adapter_CC?

revarbat:
So that the differenced parts would be slightly larger, and it would stop Z-fighting.

PR #525, opened 2021-05-05
Fix small bug in hull3d_faces

RonaldoCMP:
The bug happens with some 4 or 5 point lists like for instance:

pts = [ [0,0,0], [10,0,0], [10,10,0], [10,10,10]];

for which hull3d_faces returns: [0, 1, 3] as the points were coplanar.


RonaldoCMP:
A new boolean arg. of hul2d_path when true allows it to find not only the extreme vertices of the convex hull but also include all points that lie on the convex hull edges,

PR #600, opened 2021-07-02
Doc and formating tweaks, minor code bugs/changes

RonaldoCMP:


arrays.scad
```diff
@@ -63,8 +63,8 @@ function _same_type(a,b, depth) =
 //   a list of indices or a range.
 // Usage:
 //   item = select(list, start);
-//   item = select(list, RANGE);
-//   item = select(list, INDEXLIST);
+//   item = select(list, [s:d:e]);
+//   item = select(list, [i0,i1...,ik]);
```
revarbat, line 67: This is a case where the syntax for optional args collides with actual range or list declaration syntax.
It seems that using <optional> was confusing for some users, so we reverted back to [optional].
That is why I used RANGE and INDEXLIST instead of list or range delarations in the usage lines.

PR #601, opened 2021-07-04
rot_resample, splinesteps tweaks

adrianVmariano:
Removed unnecessary requirement for splinesteps>=3, added some error checks on splinesteps, added example of using splinesteps=1 to chamfer in rounded_prism.

Added rot_resample, rot_inverse.

### DIFF
--- a/affine.scad
+++ b/affine.scad
@@ -185,7 +185,7 @@ function apply(transform,points) =
 
 // Function: rot_decode()
 // Usage:
-//   info = rot_decode(rotation); // Returns: [angle,axis,cp,translation]
+//   info = rot_decode(rotation,[long]); // Returns: [angle,axis,cp,translation]
 // Topics: Affine, Matrices, Transforms
 // Description:
 //   Given an input 3D rigid transformation operator (one composed of just rotations and translations) represented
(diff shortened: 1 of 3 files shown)


affine.scad
```diff
@@ -230,8 +235,238 @@ function rot_decode(M) =
...
+    is_vector(beta) ? [for(entry=beta) _closest_angle(alpha,entry)]
+  : beta-alpha > 180 ? beta - ceil((beta-alpha-180)/360) * 360 
+  : beta-alpha < -180 ? beta + ceil((alpha-beta-180)/360) * 360
+  : beta;
+
+
+// Smooth data with N point moving average.  If angle=true handles data as angles.
+// If closed=true assumes last point is adjacent to the first one.
+// If closed=false pads data with left/right value (probably wrong behavior...should do linear interp)
+function _smooth(data,len,closed=false,angle=false) =
```
revarbat, line 272: This name seems likely to result in namespace collision with a function in an end user's code.

adrianVmariano: Even hidden with the underscore?   Isn't our convention that we rule the underscore namespace?  

I actually wondered if that function should be exposed...but wasn't feeling like figuring it out.  You can rename it to something more verbose if it seems necessary.   It's a general moving average that handles angles, but maybe not the best with end effects.   And I think it fails if length is even.

revarbat: You have a point about the _ prefix.

PR #662, opened 2021-09-28
path intersection speedups and misc fixes

adrianVmariano:


RonaldoCMP:
1. To avoid a division by zero in `is_path_simple` when the path has duplicate points, the test:
 `        if (/*approx(normv1,0) || approx(normv2,0) ||*/ approx(v1*v2/normv1/normv2,-1)) 1]  == [] `
should be read as:
`         if ( (normv1>eps) && (normv2>eps) && approx(v1*v2,-normv1*normv2,eps=eps))  1]  == [] `

2. Shouldn't the calls to `_path_region_intersections` in `is_region_simple` include a parameter `closed=true` ?

3. I am not sure what exactly OpenSCAD does in the iteration of `s2` but to be sure I would compute the list of segments `s2` of the region before the loop of `si` to avoid recomputing `close_region` and `pair` for each path segment.

adrianVmariano:
I think division by zero is harmless.  You get inf which is not -1.  The question is whether polygons with repeated vertices should be counted as simple or not.  Deduplicate gets run on a lot of inputs...so duplicate vertices are maybe harmless?  

Yes, closed=true is needed.

That code where s2 is computed in the loop was based on your code---originally just one for loop.  I was wondering a bit about whether that gets recomputed.   I wrote a version that didn't use pair() at all, or close_region() and it was slower, so it probably isn't doing something inefficient.

PR #944, opened 2022-09-12
Fix stroke bug (bad polyhedron)

adrianVmariano:


adrianVmariano:
Make sure the fix to stroke() looks OK and doesn't have undesirable side effects somehow.

PR #1099, opened 2023-03-27
Synopses for shapes3d, shapes2d, attachments and color

RAMilewski:
Please let me know if I'm not on the right track here.


drawing.scad
```diff
@@ -1128,6 +1140,10 @@ function _turtle_command(command, parm, parm2, state, index) =
 // Section: Debugging polygons
 
 // Module: debug_polygon()
+// Synopsis: Draws an annotated polygon.
+// Topics: Shapes (2D)
+// See Also: polygon()
```
revarbat, line 1145: Unfortunately you can't add `polygon()` to See Also, as `polygon()` is an OpenSCAD built-in which has no documentation in the BOSL2 docs.

PR #1153, opened 2023-05-01
add synopsis to coords and linalg

adrianVmariano:


coords.scad
```diff
@@ -100,6 +105,7 @@ function point4d(p, fill=0) = assert(is_list(p))
 //   pt = path4d(points, [fill]);
 // Topics: Coordinates, Points, Paths
 // See Also: point2d(), path2d(), point3d(), path3d(), point4d()
+// Convert a path to 4d.
```
revarbat, line 108: You need the Synopsis: keyword here.

linalg.scad
```diff
@@ -415,6 +427,7 @@ function hstack(M1, M2, M3) =
 //    bmat = block_matrix([[M11, M12,...],[M21, M22,...], ... ]);
 // Topics: Matrices, Arrays
 // See Also: column(), submatrix()
+// Make a new matrix from a block of matrices.
```
revarbat, line 430: Synopsis: keyword needed.

Issue #1262, opened 2023-08-22
show_geom() function?


GreenEllipsis:
**Is your feature request related to a problem? Please describe.**
I just stumbled across your library while searching for an OpenSCAD version of PCO1810 beverage cap, and I am completely on-board. My problem is that I need my cap to have a specific diameter, but pco1810_cap() has a wall parameter instead of a r|d parameter.

**Describe alternatives you've considered**
I can think of several ways to solve my immediate problem.
1. The easy out. From the function I see that my cap diameter will always be
` w = cap_id + 2*wall;`
where
`cap_id = 28.58;`
So I can calculate the wall thickness I need to achieve a given diameter. This is an unsatisfying solution.

2. Modifying pco1810_cap() to accept an r|d parameter. Pretty straightforward.

3. Introspection. I tried making a thin-walled dummy pco1810_cap(), getting the resulting radius from $parent_geom and calculating the wall I need for a given radius, and then re-rendering it in place with a child, effectively masking the dummy. This worked--almost. I hit a wall when it came to anchoring, because the object's anchoring is based on the dummy parent. Since I don't know how big the eventual child will be, I don't know how to adjust the position of the parent to compensate.

```C
module pco1810_cap_r(r, anchor, spin) {
    dummy_wall = 0.1;
    adjusted_wall = function () dummy_wall + r-$parent_geom[1];
    pco1810_cap(wall=dummy_wall, anchor=anchor, spin=spin) {
        position(TOP) %pco1810_cap(wall=adjusted_wall(), anchor=TOP, spin=spin)
            children();
    }
}

pco1810_cap_r(20, anchor=TOP+LEFT) ; 
```

**Describe the solution you'd like**
It seems like attach_geom() is partway to a solution, but won't work for pco1810_cap().
If there were a function like pco1810_cap_geom() to return geometry without rendering, I could calculate the required wall before calling pco1810_cap():

```C
dummy_wall = 0.1;
desired_r=20;
dummy_geom = pco1810_cap_geom(dummy_wall) ; 
adjusted_wall = dummy_wall + desired_r - dummy_geom[1];
pco1810_cap_geom(adjusted_wall, anchor=TOP+LEFT) ; 
```

PR #1263, opened 2023-08-23
added r|d parameters to pco1810_cap()

GreenEllipsis:
Regarding Issue #1262, I went straight forward. But still wondering about introspection.


GreenEllipsis:
You might hold up on merging this. I have the same issue with height.

GreenEllipsis:
Will submit new pull with h parameter as well.

PR #1266, opened 2023-08-29
Handles library file

GreenEllipsis:
I think this is the bare minimum to qualify as a BOSL2 library.  I ran out of time to go further.


revarbat:
This has lingered because I've not had time yet to investigate how to clean up this submission.

The top-level variables are a weird construct that isn't useful in the library, and they should be removed.
This needs some general restructuring and cleanup

GreenEllipsis:
I'm smarter on BOSL2 now than I was before. I'll give it another go as soon as I can.

PR #1308, opened 2023-11-03
Added catenary_path().

revarbat:


adrianVmariano:
Shouldn't it be called just "catenary()"?  We have "helix()" and "arc()" not "helix_path()" and "arc_path()".

revarbat:
Valid argument.  Renaming.

PR #1316, opened 2023-11-25
Added flanged bearings

MatthewShotton:
This adds support for flanged bearings, and adds a number of standard flanged bearing trade sizes.


adrianVmariano:
Maybe it should be an error to specify flange dimensions if flange==false?

adrianVmariano:
Also I think we need some additional examples that show flanges and shields.

MatthewShotton:
I've added a few more examples and some more common sizes for flanged and unshielded bearings.

I've not done much error handling in OpenSCAD before, so not sure I've landed on the best approach, LMK.

PR #1325, opened 2023-12-09
Hinges.scad - Fix snap_lock() and snap_socket()

RAMilewski:
Removed fixed $fn references in these 2 modules.

### DIFF
--- a/hinges.scad
+++ b/hinges.scad
@@ -508,8 +508,8 @@ module snap_lock(thick, snaplen=5, snapdiam=5, layerheight=0.2, foldangle=90, hi
     attachable(anchor,spin,orient, size=size) {
         back(snap_x) {
             cube([snaplen, snapdiam, snapdiam/2+thick], anchor=BOT) {
-                attach(TOP) xcyl(l=snaplen, d=snapdiam, $fn=16);
-                attach(TOP) xcopies(snaplen-snapdiam/4/3) xscale(0.333) sphere(d=snapdiam*0.8, $fn=12);
+                attach(TOP) xcyl(l=snaplen, d=snapdiam);
+                attach(TOP) xcopies(snaplen-snapdiam/4/3) xscale(0.333) sphere(d=snapdiam*0.8);
             }
         }
         children();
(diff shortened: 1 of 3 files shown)


hinges.scad
```diff
@@ -508,8 +508,8 @@ module snap_lock(thick, snaplen=5, snapdiam=5, layerheight=0.2, foldangle=90, hi
     attachable(anchor,spin,orient, size=size) {
         back(snap_x) {
             cube([snaplen, snapdiam, snapdiam/2+thick], anchor=BOT) {
-                attach(TOP) xcyl(l=snaplen, d=snapdiam, $fn=16);
-                attach(TOP) xcopies(snaplen-snapdiam/4/3) xscale(0.333) sphere(d=snapdiam*0.8, $fn=12);
+                attach(TOP) xcyl(l=snaplen, d=snapdiam);
```
revarbat, line 511: My only caveat here is that we may want to enforce a *minimum* `$fn`.  Something like `$fn=max(16,segs(snapdiam/2))`.

revarbat: And possibly we want $fn to be a multiple of 4.

RAMilewski: OK, will do.  Thanks!

RAMilewski:
I'm guessing that because we're enforcing a minimum it's ok to round down to a multiple of 4.   If you prefer, I can change it to round up.

hinges.scad
revarbat: Sorry for the delay.  I looked at the changes on my phone, but didn't want to write a comment on the phone's keyboard, and then I forgot it until now. 😞 

In any case, I just wanted to note the existence of the `quantup()`, `quantdn()`, and `quant()` functions for quantizing numbers a bit cleaner.

revarbat:
I don't actually care which way you want to round it.  `quant()` will round to the closer quantized value, though.

RAMilewski:
I really need to spend some time reading math.scad.  That's one of my weak points.  quant() is a much cleaner solution!
Part of my problem is I get seduced by the graphics in the module examples, and spend lots of time in those sections.

PR #1402, opened 2024-03-07
Update shapes3d.scad

RAMilewski:
Fix heightfield() to require bottom < 0.

Fixed circular_heightfield() base argument docs.


RAMilewski:
Missed an edge case.

PR #1418, opened 2024-04-18
Prune undefined values thru struct_set

jcorbin:
Avoid adding `[key, undef]` pairs to structs.

### DIFF
--- a/structs.scad
+++ b/structs.scad
@@ -50,7 +50,10 @@ function struct_set(struct, key, value, grow=true) =
   :
   assert(is_list(key) && len(key)%2==0, "[key,value] pair list is not a list or has an odd length")
   let(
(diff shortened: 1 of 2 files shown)


adrianVmariano:
Two notes.  One is that this functionality sort of exists via the _remove_undefined_vals() private function.  The second is that there might be a reason for a user to want a key with an undef value.  Like maybe you want to define a key as undefined but have it as a key so that a subsequent set operation with grow=false can still succeed.   Some of the existing code in rounding.scad actually works this way, but it creates the struct directly so the change wouldn't break it.  However, I do not know for sure that the proposed change is safe for the existing code.  

We could expose an option for struct_set that skips undefs.  But I'm not sure that deleting fields because they are set to undef is the right behavior.    This could all become irrelevant if native structures ever see the light of day.

revarbat:
This seems abandoned.

PR #1439, opened 2024-06-03
rect: enhanced rounding and chamfer options

hvegh:
Introduces the extra "cswap" parameter that moves the chamfer and rounding center points.

In the original code the negative roudings and chamfers are only extended in the x direction, this patch allows control over both directions.

Example:

$fn = 32;
translate([0,0,-.51])
distribute(spacing=1.5) {

// rounded examples:

    // original
    rect(rounding=.25*[1,1,-1,-1]);

    // cswap curve for quadrants [-1,1] and [-1,-1]
    rect(rounding=.25*[1,1,-1,-1], cswap = [0,1,1,0]);

// chamfered examples:

    // original
    rect(chamfer=.25*[1,1,-1,-1]);

    // cswap chamfer for quadrants [-1,1] and [-1,-1]
    // note only affects the negative chamfers
    rect(chamfer=.25*[1,1,-1,-1], cswap = [0,1,1,0]);
}
rect_enhance

### DIFF
--- a/shapes2d.scad
+++ b/shapes2d.scad
@@ -154,6 +154,8 @@ module rect(size=1, rounding=0, atype="box", chamfer=0, anchor=CENTER, spin=0, c
     }
 }
 
+
+
 function rect(size=1, rounding=0, chamfer=0, atype="box", anchor=CENTER, spin=0, _return_override, cswap = [0,0,0,0]) =
     assert(is_num(size)     || is_vector(size,2))
     assert(is_num(chamfer)  || is_vector(chamfer,4))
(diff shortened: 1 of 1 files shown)


adrianVmariano:
I don't like the cswap parameter name.  Not sure what it should be.  corner_flip?

hvegh:
> I don't like the cswap parameter name. Not sure what it should be. corner_flip?

I am horrible at thinking of sensible names, corner_flip sounds perfect. 
If that is the consensus, I am happy to change it I am happy.

Also note that a corner flip for a specific quadrant is implemented as a number: zero -> no_flip  /  non zero -> flip.

I was wondering if this should be boolean type, any preference?

adrianVmariano:
Coming up with names can be pretty difficult.  We sometimes struggle.  Go ahead with corner_flip.  

The corner_flip parameter should definitely be boolean, and it should be permitted to give a list or give it as a single boolean, so true expands into [true,true,true,true].  

Also did you confirm that the anchor overrides work properly for the case when atype=="perim"?

hvegh:
Update:

- fixed anchor overrides when atype = "perim" 
- renamed to corner_flip
- added some comments

```
flip    = true;
chamfer = 5*[0,-1,1,0];
rounding= 5*[-1,0,0,1];

distribute(spacing=40) {
    rect(size=20, atype="perim", chamfer = chamfer, rounding=rounding, corner_flip = flip)
        show_anchors();

    rect(size=20, atype="perim", chamfer = -chamfer, rounding=-rounding, corner_flip = !flip)
        show_anchors();

    rect(size=20, atype="perim", chamfer = -chamfer, rounding=-rounding, corner_flip = flip)
        show_anchors();
}
```
rect_test_perim

PR #1480, opened 2024-09-29
Propagate information about partition() part to children()

zouppen:
This patch allows to customize the partitioned parts. The part is available in `$partition_part` variable and is either `FIRST` or `SECOND`).

### DIFF
--- a/constants.scad
+++ b/constants.scad
@@ -257,4 +257,14 @@ RAY = [true, false];
 LINE = [false, false];
 
 
+// Constant: FIRST
+// Synopsis: A constant specifying the first part in partition(). `1`.
+FIRST = 1;
+
+
+// Constant: SECOND
+// Synopsis: A constant specifying the second part in partition(). `2`.
+SECOND = 2;
+
+
 // vim: expandtab tabstop=4 shiftwidth=4 softtabstop=4 nowrap
--- a/partitions.scad
+++ b/partitions.scad
@@ -566,12 +566,14 @@ module partition(size=100, spread=10, cutsize=10, cutpath="jigsaw", gap=0, spin=
     rsize = v_abs(rot(spin,p=size));
     vec = rot(spin,p=BACK)*spread/2;
     move(vec) {
+        $partition_part = FIRST;
         intersection() {
             children();
             partition_mask(l=rsize.x, w=rsize.y, h=rsize.z, cutsize=cutsize, cutpath=cutpath, gap=gap, spin=spin);
         }
     }
     move(-vec) {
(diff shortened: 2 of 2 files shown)


adrianVmariano:
Probably the values should be 0 and 1 instead of FIRST or SECOND.   I also wonder if it should be $idx or $partition_idx.  Should we rename some of the $idx variables to $<module>_idx to keep them from stepping on each other?

zouppen:
One of the hard things in computing, whether to start from 0 or 1. My initial thought was that if we ever have "BOTH" then it could be zero, for example in case we'd add an argument like `partition(keep=[BOTH|FIRST|SECOND])`.

On the prefixing question I have no preference since I'm not too familiar with your coding conventions. Of course non-colliding variable names are better than colliding :)

adrianVmariano:
In a whole bunch of places modules propagate `$idx` with a zero base to refer to iterates of the module.  See for example everything in distributors.scad.  So the simplest more clearly consistent behavior would be to do the same.  Doing `$partition_idx` could, like I said, keep things more separated, though probably a collision is unlikely in this case.  Is "both" likely?  I have no idea.  I do not like introducing FIRST and SECOND which seem like needless namespace clutter.

zouppen:
I also have mixed feelings about them, maybe my history on statically typed languages led me to introduce them. Should we use a boolean instead? I can fix the patch, but I'd like to have your opinion about naming and typing, like:

* `$partition_is_first` (true/false)
* `$partition_idx` (0 or 1)
* something else?

adrianVmariano:
Do you think it is likely that someone would want to condition on which partition within the code inside other stuff, like grid() or xcopies() or the like?  If this seems very unlikely I think maybe using `$idx` set to 0/1 would be best because it follows a pattern used in many other modules.

If we think collision is more likely then I like `$partition_idx`.

zouppen:
Hmm, maybe then I'd say it's best to go with the current convention. The collision is rather easy to avoid in user code by doing something like `$partition_idx = $idx` right inside the nested block before `xcopies()` or the like, if they need the partition index all the way down.

adrianVmariano:
Yes, exactly.  That's why I'd only deviate from that convention if it seems like a common or likely collision situation.   So I think going with `$idx` set to 0/1 makes sense.

revarbat:
Agreed with using `$idx` with values of `0` and `1` instead of constants.

zouppen:
GitHub has lost track of my commit for this pull request, maybe because I've done both rebase and force-with-lease to my branch. I close this and open a new one #1487

PR #1484, opened 2024-10-02
Update attachments.scad, fix tag_scope example

chotchki:
tag_scope only works if its used inside attachable which is the opposite of the documentation example

### DIFF
--- a/attachments.scad
+++ b/attachments.scad
@@ -1246,11 +1246,11 @@ module default_tag(tag,do_tag=true)
 //   Note that if you directly set the `$tag` variable then tag scoping will not work correctly.
 // Side Effects:
 //   `$tag_prefix` is set to the value of `scope=` if given, otherwise is set to a random string.
-// Example: In this example the ring module uses "remove" tags which will conflict with use of the same tags by the parent.
+// Example: In this example the ring module uses "remove" tags which will conflict with use of the same tags by the parent without tag_scope.
 //   module ring(r,h,w=1,anchor,spin,orient)
 //   {
-//     tag_scope("ringscope")
 //       attachable(anchor,spin,orient,r=r,h=h){
+//         tag_scope("ringscope")
 //         diff()
 //           cyl(r=r,h=h)
 //             tag("remove") cyl(r=r-w,h=h+1);
--- a/attachments.scad
+++ b/attachments.scad
@@ -1249,8 +1249,8 @@ module default_tag(tag,do_tag=true)
 // Example: In this example the ring module uses "remove" tags which will conflict with use of the same tags by the parent without tag_scope.
 //   module ring(r,h,w=1,anchor,spin,orient)
 //   {
+//       tag_scope("ringscope")    
 //       attachable(anchor,spin,orient,r=r,h=h){
-//         tag_scope("ringscope")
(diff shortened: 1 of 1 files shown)


adrianVmariano:
Why do you think tag_scope works only inside attachable?  I can't reproduce that behavior and it's not supposed to work that way.

chotchki:
@adrianVmariano Here's an example of tag_scope() not working in ways I expected which spawned the pull request. I may also be using it completely wrong (and happy to help improve documentation if so)! Where I have tag_scope now makes the most sense to me.

`
include <BOSL2/std.scad>

module outer_ring(){
    //tag_scope()
    attachable(anchor=CENTER, spin=0, d=60, l=60) {
        tag_scope()
        diff()
        cyl(d=60, l=60){
            tag("remove")
            cyl(d=30, l=61);
        }
        children();
    }
}

diff()
outer_ring(){
    cyl(d=20, l=61){
        tag("remove")
        cyl(d=10, l=62);
    }
}
`

If you toggle the tag_scope(s) you'll see inconsistent behavior. I'm using OpenSCAD 2024.09.06 (git dbd7d4f3d) with Manifold enabled.

adrianVmariano:
I took a look at your example.  I think the difference between the example in the docs and yours has to do with the children of the defined module.   If you put the tag_scope inside attachable() then it only applies to the actual object itself.  If you put it outside it will apply to the children as well.  

I'm presently uncertain if there are specific cases that call for each approach, or using it so that it applies to the children is always a mistake.  The single use in the library is for the tripod mount and it is applied your way, inside attachable().  

In the example in the docs, it makes no difference because ring() is never given any children.  If you would like to try to cook up examples that might clarify this that would be great.  That is, would you ever want the scope to include the children, and if so, when?  At the very least we should add your example, since it shows a case where it only works inside.  

The cutcube() example in the tutorial has the scope including the children, but it's unavoidable in that case, because the children are inside the diff().

chotchki:
Ah I understand a lot better now! I didn't realize the tag_scope applied to attachable ALSO meant it would apply with the same value to the "children();". I'll update this pull request to try to give a better example based on this.

adrianVmariano:
All that tag_scope does is set a $tag_prefix variable that is prefixed to all tags used, so anything in the scope (hence the name) gets that prefix.  So if you write tag_scope() {  ..... }  whatever is in the braces will get the prefix.  But in the case of attachable, we have tag_scope() attachable() { .... } so everything in those braces gets the prefix.  

I'm not sure how many distinct use cases exist for tag_scope.   Here's the one from the tutorial, which seems sort of different.  More than one example may be warranted.  

```
module cutcube(anchor=CENTER,spin=0,orient=UP)
{
   tag_scope(){
     diff()
       cuboid(15, rounding=2, anchor=anchor,spin=spin,orient=orient){
         tag("remove")attach(TOP)cuboid(5);
         children();
       }
   }
}
diff()
cutcube()
  tag("remove")attach(RIGHT) cyl(d=2,h=8);
```

adrianVmariano:
The example as written was very long and had blank lines which the example parser doesn't like.  I reworked it into a series of 3 shorter examples showing the right way and two failure modes and put that in my new PR.

Issue #1514, opened 2024-11-29
Apply textures to surface defined by point grid


amatulic:
**Problem description**
I can use `skin()` to create a skin between any two arbitrary polygon shapes. I can use textures with `linear_sweep()`, 'rotate_sweep()`, etc. But I just cannot figure out how to apply a texture to a skin.

I feel like the infrastructure for this feature already exists, and maybe it's possible to do with the existing BOSL2 functionality, but I cannot figure it out.

For example, here's a skin between a rounded square and a circle:
````
include <BOSL2/std.scad>
skin([base, rim], z=[0,30], slices=10);
````
`skin()` doesn't take a texture parameter.

**Solution I'd like**
Basically I am looking for a version of `linear_sweep()` that transitions between two polygons while applying a texture.
Better still, if `skin()` would accept a texture parameter, then I could apply a texture to any surface that can be represented by a grid of [x,y,z] coordinates.

**Alternatives considered**
I considered using `sweep_attach()` to attach my texture vnf but this gets extremeley unwieldy and unreliable.

adrianVmariano:
Applying of textures assumes that the surface follows a regular pattern so that the texture can be mapped to the surface.  For skin this is not the case.  In fact sometimes sections of the surface will vanish completely to a point.  The texture algorithm requires full repetition of the square tiles that form the texture.  So I don't think we know an algorithm for applying a texture to a skin(), which means that we are unable to provide the desired feature.

amatulic:
Perhaps I erred in using skin as an analogy. I'm really referring to a regular grid of 3D points, like an array of paths, with each path having the same number of vertices.

I was thinking, if we have an arbitrary grid of 3D points (like an array of equal-length paths), there should be a way to apply a vnf texture to each grid square. This is effectively what `rotate_sweep()` seems to do, although it does interpolations around the grid. I'd be happy just to have one vnf texture unit in each grid square, so they are all connected together.

In the case of texturing, say, a surface constructed from a square transitioning to a circle, the square would be sampled to have the same number of vertices as the circle, and the paths in between would be linear (or even bezier) interpolations from each vertex on the square to a corresponding vertex on the circle. The sweep would be from the square to the circle, interpolating the paths in between, and applying the texture as appropriate.

[...]

PR #1517, opened 2024-12-05
added squircle to shapes2d.scad

amatulic:
I have used this shape in a couple of projects and thought it would be a useful addition to BOSL2.

A [squircle](https://en.wikipedia.org/wiki/Squircle) is a shape intermediate between a square/rectangle and a circle/ellipse. Squircles are sometimes used to make dinner plates (more area for the same radius as a circle), keyboard buttons, and smartphone icons. Old CRT television screens also resembled squircles.

The basic shape looks like this:

image

I have implemented a module and function for a squircle in `shapes2d.scad`. It takes a parameter for `squareness` (between 0 and 1, where 0=circle and 1=square), and a `size` parameter the same as with `square()` to define the bounding box for the squircle.

Because the curvature can change sharply at the corners, the vertices are distributed uniformly only when the shape is a circle (`squareness=0`). The vertices concentrate more toward the corners as `squareness` approaches 1. This results in a fairly smooth shape even with low `$fn` values.


adrianVmariano:
Is this shape different than the shape produced by supershape(), e.g. those in example 3?  The wikipedia page for "squircle" suggests that the most common "squircle" is the superellipse, and we already have that as a special case of supershape.

amatulic:
Indeed, supershape example 3 generates squircles. The difference is in implementation and capability:
* The supershape doesn't respect `$fn` although it does have a step parameter. The squircle uses `$fn` instead.
* Because the supershape can produce many different kinds of shapes with curves and cusps, it is impossible to optimize the number and distribution of vertices, instead relying on a very small step size to resolve sharp curves. The squircle is predictable and the step size is varied as necessary to produce smooth corners.
* The supershape squareness parameter is unbounded, ranging from 2 to infinity. The squircle squareness parameter ranges from 0 to 1, allowing for better control.
* Squircle parameters are simpler to understand, requiring only a squareness and bounding box `size` like a square.

The squircle wouldn't be the only thing in BOSL2 in which it's possible to do something multiple different ways. For example, you can use `circle()` or `supershape()` to produce an ellipse also, you can use either `offset()` or the rounding features of BOSL2 to make rounded polygon shapes, and you can approximate any curved shape with Bezier paths too.

The ability to have a small number of vertices also make it feasible to generate a "sphube" (intermediate between shpere and cube) as a VNF with a reasonable facet count. I was planning to do that in a future PR.

adrianVmariano:
Is there some compelling reason to not make the "squareness" parameter **exactly** linear?  That would make it more usable for somone who wants to know where the curve will end up, for example.   Wait, did you implement the FG squircle?  That's totally different than the superellipse squircle.   If that's the case then exact adherence to the FG equation could be an argument for nonlinearity.  This might be something to bounce off Revar and Richard in the chat. 

We might also want to show some kind of comparison of how this compares to round_corners with different k values or just a circular rounded square.  

The 8 anchors should lie on the boundary of the shape if you have it right, with the arrow perpendicular to the shape.  

Try https://gitter.im/revarbat/BOSL2 or if that doesn't work, try going to app.gitter.im and then see if you can find the BOSL2 chat somehow.

amatulic:
I can make it closer to linear. I could also make it nearly exact, but I would need to introduce a curve-fitted polynomial correction with hard-coded coefficients. The correction there now is a simple formula, basically

`new_squareness = sqrt(squareness*(2-squareness))`

This results in a maximum error when squareness=0.48, in which case the corner of the squircle is 0.55 of the way to the corner of the square. If I use the fourth root instead of the square root, the error is even less, with the maximum being at 0.04, making the "corner" (it's almost exactly a circle at that point) 0.036 away from the corner of the square.

I fixed the anchor problem and pushed a new commit, but the anchors don't account for the squareness correction. It's past midnight here, so I'll revisit this tomorrow.

amatulic:
Ah, that's better. Perfectly linear now! This is the demo using 10 intervals of squareness between 0 and 1:
image
I have pushed the fix. You can test it if you want.

adrianVmariano:
Ah, I was going to say it's easy to transform it to a linear parameter from the equation----no research paper required.   You just identify the desired "corner" coordinate R*[a,a], where a needs to go from 1/sqrt(2) for a circle up to 1 for a square.  If b is a linear parameter on [0,1] then we have a=lerp(1/sqrt(2),1,b).  Plug that into the FG equation and solve and you get s^2 = (2*a^2-1)/a^4.   Note that it's equally simple to transform the exponent of the supershape into a linear parameter on [0,1].  

I haven't delved into the literature.  The argument for using the FG equation squareness parameter directly would be that the optics folks who invented it know and use it with that original functional form and would be confused to encounter it with a different parameter.  I'm not saying that is compelling for our situation.  

Another question arises:  is the FG squircle the right one to use?  Wikipedia says it's the "less common" version.  (I've known about the superellipse for years and never heard of the other thing until yesterday.)  Between this one and the superellipse, does one have an advantage?   Smoother? Looks better?  They are definitely different, so I want to know why we picked FG.  Or should we support both with a squircle style flag?

amatulic:
I took out "most common" from the Wikipedia article. An assertion like that needs sources to back it up, and the sources I have found mention both types.

I suspect that it doesn't matter how you arrive at the squircle, I suspect it may be the same shape from either method. I'll have to do some tests on that.

The advantage in OpenSCAD is that it's computationally less expensive to avoid dealing with fractional exponents, but this likely wouldn't make much difference in execution time.

Should I revise this to use superellipse?

adrianVmariano:
No, it's definitely not the same shape.  I mean, that's obvious from the functional form that it can't be, but I plotted a superellipse on an FG and they were visibly different, with the superellipse being more squarish.  

You're using trig functions in the computation and worried about some fractional exponents?  I expect run time for sin and cos are comparable to pow.  Computation time should not be a concern, in any case. 

My question above was an honest one.  While to  me the mathematical formulation of the superellipse is more elegant, I don't know for practical applications, what distinguishes the two shapes or what would make the FG shape better/worse than the superellipse.  If you had both available...how would you pick?  

It would also be interesting to compare the FG supershape to a continuous curvature rounded rectangle, which because it uses a 4th degree bezier, **could** actually be the same shape.   (But actually not really because the rounded rectangle has a flat section.)

I noticed that the squircle references indicate that the superellipse version is the one specific superellipse with n=4.  

Regarding which curve is better known or more used, it seems that the superellipse has got to be the winner there.   The superellipse equation was described 100 years before FG wrote their paper.  It was used in a font in 1952.   But under the specific name "squircle"?  That name is clearly a recent coinage, apparently much newer than either the superellipse or the FG shape.

adrianVmariano:
image

amatulic:
You beat me to it. I was just starting on the comparison myself. Thanks for the plot. Yes, there's a subtle difference.

And yes you're correct, I was searching for articles containing the word "squircle", so naturally everything I found mentions the FG squircle as well as the superellipse squircle.

You asked:
>If you had both available, what would you pick?

I'd pick FG because it is mathematically well-behaved. While the expression for the superellipse is elegant in its simplicity, it's highly inelegant *mathematically* in its behavior. The main disadvantage to the superellipse is that you're dealing with an infinite exponent for squareness=1. Even when you get close to 1, the exponent gets quite large, potentially causing computational instability (although it's unlikely for expected use cases).

As I said near the beginning of this discussion, the FG squircle (quoting Wikipedia) "allows a smooth parametrization of the transition to a square from a circle, without involving infinity." I'd say that would be its main advantage.

amatulic:
Thinking about it some more, I'll retract my statement about mathematical behavior. The polar form of the FG squircle doesn't behave well either when the angle is multiples of 90° and I deal with that simply by setting the squircle radius to the circle radius in those cases.

For the superellipse exponent going to infinity, the squareness parameter could be clipped at 0.99 (I've implemented it here and it blows up past 0.993), resulting in a shape indistinguishable from a square with the same number of vertices as the squircles. Retaining the vertex count would be useful for skinning between two squircle shapes.

adrianVmariano:
I didn't ask you which you'd pick as a programmer.  I want to know which one you'd pick as a user who doesn't care how the thing is implemented, and whether it was computationally tricky or not.   It's the job of the library author to handle those challenges and do the right thing rather than the easy thing, because we're doing it once and for all.  

I'd argue that there's nothing wrong with infinity in math and there's nothing ill behaved about the superellipse.  In fact, it's quite neat.  

The characteristic you're complaining about is a computational issue, not a mathematical one, and it doesn't appear to be particularly bad.  In fact, it seems to be surprisingly robust.   It appears that maybe in the case of cartesian coordinates, the roundoff error produces errors that give you the right answer (square), for example.  In polar coordinates it may not be so tidy, but I still got a good point with an exponent of 2000.  I thought it would fail at a lower value.  It appears to me that failure is primarily a result of underflow error.   

The mathematical behavior of the curves I'd consider to be things like its derivatives, or the curvature along the curve: the properties of the curve as an ideal mathematical object.  Does the superellipse have anything to recommend it over the FG curve in terms of these properties?  Why would someone who knows nothing about the mathematics of the curves pick one or the other?  Are there mathematical properties that suggest picking one or the other?  Can we even distinguish the curves from each other at all if we see them without context in isolation?  (Quick, is this a superellipse or an FG?)

One can ask the question, why not just use a circle.  When I wrote all the continuous curvature rounding stuff someone had claimed that such rounding was what apple used for their device corners.  It seems like the story always changes.  But continuous curvatures really does look and feel different.  In a 3d printed model you don't see a highlight where the roundover starts and it feels smoother as well.   It seems likely that curvature for both FG and superellipse is continuous, though perhaps something odd happens at face centers.  

So suppose we implemented everything and now we have to write a rounding tutorial and tell a user how to pick between superellipse, FG, using rect() with circular rounded corners, or using continuous curvature bezier corners.   How does a user decide?

RAMilewski:
As a user concerned primarily with designing objects, (and one who would likely be befuddled trying to work through the underlying math), I would argue for including both forms and using a flag to make the selection.   That makes it possible to try both quickly and use the one that provides the best fit for the current model.

amatulic:
It's curious that you managed a higher exponent; was this in OpenSCAD? I wrote it separately from BOSL2 to test.

Good questions.
> The mathematical behavior of the curves I'd consider to be things like its derivatives, or the curvature along the curve: the properties of the curve as an ideal mathematical object. Does the superellipse have anything to recommend it over the FG curve in terms of these properties?

Well, in polar form, the derivative of r with respect to theta is a cleaner-looking expression for the superellipse than for the FG squircle (I tried both in Wolfram Alpha). For both shapes, the curvature continuous and smoothly changing. For both shapes, nothing odd happens at face centers because the underlying mathematical functions are differentiable and the derivatives have no discontinuities.

>Why would someone who knows nothing about the mathematics of the curves pick one or the other?

Someone may want the superellipse to duplicate shapes made with it (like fonts, certain buttons on devices). Aside from those who work in optics and already use the FG curve, someone else may want the FG version because literature on superellipse covers vastly more than the specific case of a squircle, and literature on the FG squircle covers exactly that.

The reason **I** picked FG years ago was because it was the first thing I found in polar form during a literature search for "interpolate between a circle and a square" (I didn't even know the term "squircle" at the time), and the Wikipedia article back then didn't help me figure out how to implement it in OpenSCAD, especially with all that stuff about gamma functions. I am the person who added the polar form to that article, just yesterday.

>Are there mathematical properties that suggest picking one or the other?

I don't know. From my perspective, infinite exponents are undesirable, and I know your view is different.

>Can we even distinguish the curves from each other at all if we see them without context in isolation? (Quick, is this a superellipse or an FG?)

In isolation, they are indistinguishable. Compared overlaid, the superellipse squircle is ever so slightly more rounded than the FG squircle.

>One can ask the question, why not just use a circle.

* Efficiency. For example, a squircle-shaped dinner plate (my family had these when I was a child) holds more food but still takes up the same space on the shelf as a circular plate. For use as a physical button or a smartphone display icon, a squircle can still contain text more easily than a circle. I thought Apple used them, but looking at my wife's iPhone next to my Android phone, both look like they use rounded squares nowadays.
* Aesthetics. There's beauty in smoothly curved objects that have no flat parts. I like designing objects that make smooth transitions between shapes, like propeller blades and ergonomic handles.
* Nostalgia: Before flat-screen displays became commonplace, cathode-ray-tube television screens resembled squircles. A project to design an old-fashioned TV enclosure, in fact, is what led me to discover the FG squircle and start using it, before I even knew about BOSL2.

>So suppose we implemented everything and now we have to write a rounding tutorial and tell a user how to pick between superellipse, FG, using rect() with circular rounded corners, or using continuous curvature bezier corners. How does a user decide?

Even though I prefer FG, in light of that question, the best approach to avoid confusion would be simply to not offer FG and base the squircle on the superellipse. This maintains consistency with what BOSL2 already offers.

Continuous curvature bezier corners would be more difficult to deal with if a designer wants an attractive interpolation that can go all the way from circular to sharp square corners. I use beziers in almost all my projects, but for the recent discussion in #1514 about a basket that transitions from a square to a circle, I would want a squircle for the perimeter, not beziers. I would use beziers to control how the squircle radius changes as a function of z.

As for squircle versus square with rounded corners, this may be more of an aesthetic choice depending on the rounding. The smooth roundover could simulate a squircle with the right combination of the `cut` and `k` parameters. Using roundover to make a squircle wouldn't be *my* choice; I find this a bit fiddly because the two parameters aren't independent (minimum value `k` that doesn't produce an error depends on the value of `cut`).

amatulic:
> As a user concerned primarily with designing objects, (and one who would likely be befuddled trying to work through the underlying math), I would argue for including both forms and using a flag to make the selection. That makes it possible to try both quickly and use the one that provides the best fit for the current model.

Well, the thing is, they're both visually indistinguishable unless they're overlaid. Just different approaches to create a shape. Would providing two choices to accomplish effectively the same result confuse the user?

The problem here is that I unwittingly introduced a new method to produce a squircle, when other methods to do it were already available in BOSL2, just not implemented for this specific shape with the parametric control offered by the `squircle()` function in this PR.

RAMilewski:
I would argue that adding the squircle() function would still be useful.   It's an issue of discoverability.  For a rank beginner finding the existing squircle methods would require scrolling through the examples and stumbling upon a squircle.   Nothing in either the topics index or the function index would help guide the casual user to any of the exiting methods.

amatulic:
@RAMilewski - Oh, I think we all agree it should be there. The question now is what version of squircle() should be implemented given that they are nearly indistinguishable, or whether we should make a choice with a tag as you suggest, even though the user is unlikely to be able to tell the difference.

adrianVmariano:
I did my numerical testing in MATLAB with r=1.  In the polar case, the R value factors out so that's how the implementation works anyways.  I also only spot-checked the point at 45 deg for that test so maybe it fails for some other theta at N=2000.   I did the full curve in cartesian, again for r=1.  It's actually sort of obvious why it works really well in the cartesian case:  The equation is x^n+y^n=1, which solves to y=(1-x^n)^(1/n) and when n is too big, x^n->0, but 1^n is always 1 no matter what.  So it reduces to 0 to 1/n which will be zero always  until n is so big that 1/n underflows to zero, and then 1 always in the second case.  Very tidy actually.  Substantially less tidy in polar coordinates, unfortunately.  The cartesian representation is obviously bad from a point sampling perspective.  

Note that by mathematical behavior I don't mean what the expressions look like but what they do.  Like what does the graph of curvature look like as a function of theta?   Does one of these things have a smaller curvature than the other one, or distribute its curvature in a different way?    What is the minimum/maximum curvature as a function of squareness?  

I'm not sure what you mean about someone preferring the FG because "literature on superellpse covers vastly more than the specific case of a squircle."  

There is no problem with overflow in computing the superellipse, only underflow, and as previously noted, in cartesian coordinates it seems to actually magically handle itself and producing the n=infinity case automatically, which is pretty tidy.  

Indistinguishability in isolation is not necessarily an argument that they are completely interchangeable.  It's an obvious first question to pose.   Maybe a better question is whether a user can tell the difference if they flip the flag that changes the type so they see the curve flip.  Heck, maybe we should support both only because nobody does that and it lets someone who is curious easily flip back and forth and see both forms and how different or not they are.  Nobody has to use the flag that changes the type if they don't want to.  

Note that just because the two things are similar doesn't mean we can't have both.  We can make one the default, document that they are "slightly different" and let users choose.  Maybe somebody seems them as markedly different, or maybe the exact shape of one interacts with a model in some way that works better for somebody.  

I should have been more clear when I posed the question "why not just use a circle".  What I meant was a circular rounded rectangle.  

You say that continuous curvature bezier corners would be more difficult for going from circular to sharp, which I suppose is true.  But since nobody should ever want actual sharp, one can counterargue that for example a rounding with maximum joint value and k going from 0 to 0.92 actually fits the bill quite well.   Note that if you're trying to make the maximum width rounding you shouldn't be trying to minimize k but instead should be setting joint to the amount of space you have and then tuning k to control the shape you get.   

The above shape:

image

```
klist = lerpn(0,.92,15);
p=[ for(k=klist)
      round_corners(square(10), joint=5, k=k, method="smooth")
  ];
skin(p, z=klist*6, slices=0);
```

It just occurred to me that you should make sure that the corner anchors are doing something reasonable in the case of non-square sizes.  I am thinking "reasonable" means they point out at 45 deg and are located at the corners.  I wonder if we need "box" and "perim" anchors e.g. like rect() has.  Probably the answer there is yes.

amatulic:
I did my numerical testing in OpenSCAD and couldn't get an exponent working past 236 without an error.

The graph of the curvature as a function of theta can be visualized roughly by plotting the angular difference of each line segment. I didn't do this but since both curves are differentiable (dr/dtheta) I would expect it to be smooth for both superellipse and FG until they approach a square.

>I'm not sure what you mean about someone preferring the FG because "literature on superellpse covers vastly more than the specific case of a squircle."

That was my experience. There's so much information about superellipses that the thing I was looking for (interpolation of a circle with a square) landed me on sources discussing squircles, and those included the FG squircle, which seemed more straightforward in the papers I saw than the superellipse, which I noted involved an unbounded exponent.. Now that I know more, I recognize the superellipse is simple to deal with.

>Heck, maybe we should support both only because nobody does that and it lets someone who is curious easily flip back and forth and see both forms and how different or not they are. Nobody has to use the flag that changes the type if they don't want to.

I agree.

>I should have been more clear when I posed the question "why not just use a circle". What I meant was a circular rounded rectangle.

My answer would still be similar, aesthetics and nostalgia. There is something pleasing about the shape that a rectangle with straight sides and circular rounded corners. I agree with the point of view you expressed that a 3D printed surface without flat regions is attractive.

>You say that continuous curvature bezier corners would be more difficult for going from circular to sharp, which I suppose is true. But since nobody should ever want actual sharp, one can counterargue that for example a rounding with maximum joint value and k going from 0 to 0.92 actually fits the bill quite well.

Yes, I expect so. I try not to design things with sharp corners anyway. The corner can't be sharper than the 3D printer's nozzle diameter after all.

>It just occurred to me that you should make sure that the corner anchors are doing something reasonable in the case of non-square sizes.

The anchors I set are on the corners and edges of whatever squircle you make.
image

adrianVmariano:
Here's my test recoded in OpenSCAD.  It fails somewhere above n=2150 in OpenSCAD on my machine.  

```
function shape(n,theta) =
  [for(t=theta)
       let (x=cos(t),
            y=sin(t),
            r=(x^n+y^n)^(1/n)
       )
       [x,y]/r
  ];
theta = lerpn(0,90,181);
path = shape(2150,theta);
stroke(path, width=.003);
```

Perhaps there's a complication arising in the non-square case.  I really haven't thought about that case.  It seems like from a computational point of view it should just be a rescale of the unit case.  There may be an issue about what the squareness parameter means, though.  

Note that your example is not a non-square case for displaying anchor behavior.   Wait, actually you have an issue with how anchors are done.  We don't want any named anchors, just the basic anchors.  I think we need an atype flag and two kinds of anchors, anchors on the notional square the size specifies and anchors actually on the shape like you already have but for the second kind, you need to use the anchor override feature to specify them, not named anchors.  

The question about curvature isn't whether it's smooth, which it will be since both functions are C^infinity, it's what its max and min values are.  That's one thing that could distinguish the two super shapes from each other, if say the FG has a higher curvature at its point (for same parameter) you could say it's "pointier".

amatulic:
Hmm. I'm using supershape(). I convert squareness to n for linear interpolation of the corner positions, and it fails when squareness>0.993. The size or aspect ratio have no effect. So I'm clipping it at 0.991, which is still quite sharp; you have to increase the number of points to 256 and zoom way in to see a tiny dulled corner.

As far as being pointier, you can see which one is pointier just by looking at it. Here's one quadrant:
image
The cyan one on the bottom is the superellipse, and the gold on top is the FG squircle. The FG clearly has a sharper corner. The difference disappears at squareness=0 or 1, of course. It's biggest at middle values.

I added a "type" parameter that can be set to "fg" (FG squircle) or "se" (superellipse) and added the superellipse. The picture above is from the new shapes2d.scad. I haven't pushed it yet.

I fixed the non-square case with anchors. I noticed that earlier today. I used named anchors because that's what reuleaux_polygon() uses. I just copied the code I saw and modified it a bit.

I'm going to bed now. I'll study the anchor documentation tomorrow.

adrianVmariano:
Well, supershape() is much more complicated than the superellipse so I'm not surprised it bombs at a smaller n.  

The basic anchors assume a rectangular shape and you can do some things with the basic anchors that you cannot do with named anchors because of that rectangular assumption.  The reuleaux polygon needs named anchors because it is non-rectangular.  It also supports the basic anchors---everything does automatically---but they may be wonky and not useful if your polygon doesn't have at least an even side count.    

You mentioned that you don't use anchors.  Does that mean you don't do things like `anchor=BOTTOM` or the like when creating shapes?  

I was pondering the correct parameter name because spheroid uses "style".  It seems to be the only example of a module that generates different versions of the same output, though, so it's not a very strong case.  However, you're going to need an atype parameter for anchoring, so that suggests to me that style might be better just because it's more distinguishable.   The other thing is that we make an effort to avoid cryptic abbreviations.  I think "fg" can't be avoided (probably should accept uppercase and lower case), but for the superellipse you need to use a word, so either "super" or "superellipse" for that option.  

Does the documentation need to mention the N=4 superellipse squircle and whatever the magic parameter is that gets that shape?

amatulic:
The maximum exponent you got corresponds to a squareness of about 0.99889.
I wrote this function to calculate the superellipse squircle radius:
````
function superellipse_radius(n, r, angle) =
    (abs(cos(angle)/r)^n + abs(sin(angle)/r)^n)^(-1/n);
````
... and it fails at the same point. I'll use that instead of supershape. I can clip the squareness at 0.998.

No, actually I've never used anchors. Up until now, most of my work has been outside of BOSL2 because:
* I developed my own library of things analogous to VNF, threading, and filleting long before I discovered BOSL2 about 1.5 years ago when I was looking for the download link and created #1145 as a result.
* The Thingiverse customizer requires standalone .scad modules. Makerworld now supports BOSL2 in its customizer, but I don't use Makerworld, my primary repos for my work are Thingiverse and Printables, and I wish Printables would have a customizer.
 
I didn't do much with BOSL2 until a year ago when I badly needed to use offset() and union() on my own paths, resulting in in #1312 about an issue that came up in union(), which you fixed. That project never got finished (it was for a Christmas contest deadline that I didn't meet) but I should finish it before Christmas this year. Most of my work involves extrusions and polyhedrons rather than pre-made shapes, and I had already written my own version of vnf_vertex_array() before I knew about BOSL2. The need for anchors has rarely come up in my work.

I'm not sure what you mean by anchor overrides. If there's an example, in shapes2d.scad, that would help. For now I have just removed the side anchors because they duplicate the default ones, and kept the corner named anchors.

An alternative to a parameter called "style" or "type" might be "method". I'll use "style" for consistency with spheroid.

The documentation shouldn't mention the special case of n=4. A squircle is any value n>2. When n=4, the squircle outline corresponds to the *geometric mean* (not arithmetic mean) between the circle and the square. That's the only thing special about it, and I doubt any user would care. The point here is to create a squircle with an intuitive squareness parameter that gives an expected interpolated corner position. The exponent n=4 actually corresponds to a squareness of 0.4568.

**Comment by adrianVmariano** [comment]
I see letting efficiency fall by the wayside and computing sin(angle) and cos(angle) twice, eh?  :)

You might want to check out the attachments tutorial to learn about what anchors and attachments can do.  It's one of the most powerful features in BOSL2 and really makes it a lot easier to design models that are built up from multiple parts.   (It won't help if you're making a single complicated extrusion of some sort.)   It lets you position objects relative to other objects without explicit transformation operations.   (Unfortunately it can also be a significant PITA for the programmer designing new modules for the library.)

https://github.com/BelfrySCAD/BOSL2/wiki/Tutorial-Attachments

I've said already that your anchoring should follow the method used by rect() which is in shapes2d and has two type of anchors.   Anchor override is where you replace the standard anchors with your own definitions instead of creating new named anchors with your own definitions.   What you need to do is have an atype parameter that takes a "box" and "perim" option and for box you don't use an override and you specify attachable as a rectangle with a size specification.  For the perim option you do the same but with an override, which will replace the corner anchors with anchors that are actually on the corners.  Basically this should be a drop-in replacement for rect() except for the rounding options that rect() has.  

Does it make sense to allow the squareness parameter to go below 0?  The supershape range from n=1 to n=2 is also potentially useful as the circle collapses to a diamond.  A similar thing happens with FG, though I think it never actually gets flat sides.  I'm not sure and haven't made the effort to figure it out.  

The sole reason I singled out the special case of N=4 is that multiple references claim that "squircle" refers to specifically and only the superellipse with N=4.  That one single case, not any other N values and not the family of superellipses between the square and circle.  Someone therefore might want to be able to make "the" squircle, and would therefore need to know the parameter to do so, since it isn't obvious.    

(Note: I don't understand what's meant by the (geometric) mean between two curves.)

**Comment by amatulic** [comment]
> I see letting efficiency fall by the wayside and computing sin(angle) and cos(angle) twice, eh? :)

Oops! Good catch. Stupid of me. Fixed on my end.

Thinking back on my projects, I have not done much attaching small parts together. My more complex designs usually involve complex polyhedrons, and may have parts that need to fit together after printing, but these are designed separately. I do have an "assembly view" option in my models that would lend themselves to attachments, however.

> Does it make sense to allow the squareness parameter to go below 0?

I'm not sure. It wouldn't work with FG, though, because the squareness term is squared in the equation, making it always positive.

> The sole reason I singled out the special case of N=4 is that multiple references claim that "squircle" refers to specifically and only the superellipse with N=4 ... Someone therefore might want to be able to make "the" squircle

That's a fair point. In that case I'll mention the n=4 case in the documentation, giving the equivalent squareness term.

> (Note: I don't understand what's meant by the (geometric) mean between two curves.)

It's the geometric mean betwen the two radii. Say the circle has radius rc=1, then square at the corner has radius rs=sqrt(2). The geometric mean of n values is (a1*a2*a3...an)^(1/n). The geometric mean between the the circle and the square corner is (rc*rs)^(1/2), which is 1.189. This is 0.456786 of the distance between 1 and sqrt(2), therefore the equivalent squareness parameter is 0.456786.

**Comment by adrianVmariano** [comment]
"It's the geometric mean between the two radii."

That was sufficient detail.  Just FYI, I am a mathematician.  Let that guide your level of detail in mathematical communications.  :)

Going sub-circle works fine with FG.  You just flip the sign.  But actually there's no reason the squareness factor needs to be squared anyways if you're actually solving for it to linearize.  In fact...why would you square it given that?   Why did they square it to begin with, actually?   I mean, the basic formulation appears to be an interpolation between an equation for the circle and an equation for the square.  I mean, did anybody say s had to be real valued?  Imaginary values are subcircle...

Note, you might actually run some timing tests to see if it matters.  I predict that running sin and cos and extra time will be irrelevant even with 1000 points.  But you can see if I'm right or not.  I find it's often hard to predict run times for OpenSCAD code.  I mean, generally list comprehensions are best, tail recursion is next, I think C-style list comprehensions may be next.  They are slow.  And non-tail recursion.  Of course the other major issue with non-tail recursion is blowing the stack, which is easy to do.

**Comment by amatulic** [comment]
I am not a mathematician. My background was physics but I was always a bit weak on the math even though I'm familiar with it.

Calling math functions unnecessarily twice is just bad practice, even if it doesn't matter in OpenSCAD, I try to avoid it.

I'm using the polar form of the FG squircle. If you negate squareness, it just makes radius negative. If you flip the sign under the radical, you get square roots of -1.

I'm still trying to figure out the anchoring override.

**Comment by adrianVmariano** [comment]
Well, in programming sometimes ease of code maintenance or comprehension is more important than efficiency these days, so it's important to find the balance.   On my computer a sin() evaluation is taking 0.7 microseconds.  You need a LOT of those before it matters.  And it's almost always the case in OpenSCAD that something orders of magnitude slower is sitting on top.  So if it's more readable, more elegant, or more maintainable, then call the function twice.  

I'm not sure what you're talking about regarding the radius being negative.  Here's the s>1 case and the s<1 case where I took s to be an unsquared parameter and I've set the radius to 1:

image

So blue is a subcircle and red is a standard squircle.    It's actually obvious that the argument of the square root is always positive when s<0 because in that case it's the sum of 1 and a nonnegative term.  It's only for the *regular* squircle where this is not immediately obvious.  

image

But even there, things are well behaved as seen above for larger values of s.   A more careful analysis shows the minimum of the term under the square root is always at 45 deg angle points.  At 45 deg we have 1-s*sin^2(45)cos^2(45) = 1-s/4, so we'll hit zero when s=4, meaning that we're valid for all in (-infinity,4), though we'll get cross arms reaching out to infinity as s->4.   

Here's the shape you get if you solve for the 45 deg angle points being on the diamond:

image

If you understood named anchors you can do anchor overrides the same way.  Just instead of a name, give the anchor direction, like [1,1] for the back right corner.  I guess there's no helper function so you have to list out the anchor components yourself in a list instead of passing them to a function.

**Comment by amatulic** [comment]
> Just instead of a name, give the anchor direction, like [1,1] for the back right corner.

That helped clear things up. The problem is, the anchors always appear on the perimeter.

The all the squircle() module does is do some error checks, call the squircle() function, and attach the anchors:
````
module squircle(squareness=0.5, size=[10,10], style="fg", atype="box", anchor=CENTER, spin=0) {
    check = assert(squareness >= 0 && squareness <= 1);
    anchorchk = assert(in_list(atype, ["box", "perim"]));
    bbox = is_num(size) ? [size,size] : point2d(size);
    assert(all_positive(bbox), "All components of size must be positive.");
    path = squircle(squareness, bbox, style, atype, anchor, spin);
    attachable(anchor, spin, two_d=true, path=path) {
        polygon(path);
        children();
    }
}
````
Tne squircle() function does all the work. It does some checks, calls functions to generate either a FG or superellipse path, then sets anchor corner coordinates to the bounding box coordinates if atype="box" or to the squircle corners if atype="perim". An override array is set to `undef` for atype="box" or the anchor positions if atype="perim". But show_anchors() always shows anchors on the perimeter.
````
function squircle(squareness=0.5, size=[10,10], style="fg", atype="box", anchor=CENTER, spin=0) =
    assert(squareness >= 0 && squareness <= 1)
    assert(is_num(size)     || is_vector(size,2))
    assert(in_list(atype, ["box", "perim"]))
    let(
        path =
            style == "fg" ? _squircle_fg(squareness, size) :
            style == "superellipse" ? _squircle_se(squareness, size) :
                assert(false, "Style must be \"fg\" or \"superellipse\""),
        bbox = is_num(size) ? [size,size] : point2d(size),
        rx = 0.5 * bbox[0],
        ry = 0.5 * bbox[1],
        corners = atype == "box" ? [ [rx,ry], [rx,-ry], [-rx,-ry], [rx,-ry] ]
        : [ for(i=[1:2:8])
            let(ca = i*45,
                cp = polar_to_xy(_squircle_anchor_radius(squareness, ca, style), ca)) [rx*cp[0], ry*cp[1]]
        ],
        anchorpos = [[1,1],[-1,1],[-1,-1],[1,-1]],
        override = atype=="box" ? undef
            : [ for(i=[0:3]) [anchorpos[i], corners[i]] ],
      ) reorient(anchor, spin, two_d=true, size=bbox, p=path, override=override);
````
Do you see anything obvious that I missed?

**Comment by adrianVmariano** [comment]
I wouldn't call it something obvious, but something subtle.   First of all, functions call reorient which lets you use anchors for changing the position of the object.  This just changes the output of the function by translating and possibly rotating the path that you have computed.  If you want to tell whether those are working you need to do things like

polygon(squircle(...., anchor=[1,1]));

That should put the BACK+RIGHT anchor on the origin.  

When you call squircle with show_anchors() you are of course using the module form.  Anchoring for modules is accomplished with attachable() instead of reorient().  You need to provide the original centered shape to attachable, so you have to call the function form WITHOUT an anchor, spin (or in 3d, orient).  If you scrutinize rect() you'll see that those things are not passed when the module calls the function.  

Unless you want to duplicate code you need to create a mechanism for your function form to pass override back.  Again, take a look at rect().  It's got a private option, signaled by leading underscore, to make it return the override.   You then invoke attachable with the override.  

There's one other thing that is wrong that is perhaps even more subtle.  You have given both attachable and reorient two different pieces of "size" information.  This should be an error according to me.  According to you it should pick one and do something.  Well, it did the second thing.   I don't know which one it picks, so maybe it will work, maybe it won't.  What I mean is that you have given it both a size and a path.  When you specify a size it makes rectangular anchors based on the sides and corners of the rectangle.  When you specify a path it assumes you've given it some arbitrary shape without structure and it makes anchors by using one of two algorithms on the shape that I won't detail right now for trying to decide on reasonable right, back, etc locations and directions.    Hmmm....code inspection indicates that you're in luck and it ignores path if you've given size.   But nevertheless, I advise not passing path.  You'll see that rect() doesn't.

**Comment by amatulic** [comment]
Got it. It's working now. I was initially confused by what at first appeared to be rect() recursively calling itself, and then I got tripped up by the override aray needing nested coordinates like [ [[x0,y0]], [[x1,y1]]...], but once I figured that out, the anchors displayed properly.

I wasn't giving attachable any size information in the code above. No error appeared. I'm passing the size to it now, though.

Now I have to figure out why the displayed anchors are repositioned correctly when I set different anchor coordinates, but the shape is being shifted by its diameter rather than its radius.

**Comment by adrianVmariano** [comment]
You are still passing both size and path to reorient().  

The module version is passing anchor and spin to the function version, so the anchor offset will get double applied.  You need the module version to use the centered version of the shape, so don't pass anchor or spin.  The anchor and spin will get handled by attachable().  

Note as a style point, colons should be the first character on a line, not the last one.  The colon is like the "else" that makes it clear a new block is starting, and it's kind of hidden when you put it at the end of the line.   So write a conditional chain like:

```
foo = condition1 ? result 1
    : condition2 ? result2
    : condition 3 ? result3
    : result4;
```
(You'll see this convention generally followed throughout BOSL2, though not universally because we didn't figure it out at first.)  

Also instead of putting tests for "box" inside the loops do it outside.  Set corners to [] or undef if atype is box and set override to [] if atype is box.  

I don't understand your last statement in the "however".  Your approach should be identical to rect.  I think if you fix the problem of passing the anchor in when you call the rect function from the module that should fix it.

**Comment by amatulic** [comment]
I thought I was doing it identical to rect, setting override to an array of undefs like that instead of just having override be undef, passing size in both the function and the module, etc. I'll look through it again. Once I get it working, I'll push another commit.

**Comment by adrianVmariano** [comment]
Uh, I guess I keep saying do exactly what rect() does, but in rect() each corner is treated individually and may or may not be rounded, so you have to do the check inside the loop.   In your case, it's all or nothing, so avoid the complication and do the check once outside the loop.  

And you're right: reorient needs to have the path because it returns the path, so I was wrong about that.  It's just attachable that doesn't need the path and I think you had that right.

**Comment by amatulic** [comment]
I finally saw it. The module doesn't pass anchor and spin to the function when it overrides. I missed that. I removed them and now everything works. Testing with anchor, spin, various rectangle sizes, everything is working now, both with module and with polygon(). Doing some code cleanup now.

**Comment by adrianVmariano** [comment]
Ok.  Let me know when you are happy with it and have all the examples in place and I'll "approve" it, which will run the tests and if it passes, then Revar can merge it.  (Revar has been dealing with real life issues that have limited his time on this project, so he's sometimes a little slow to merge stuff these days.)

**Comment by adrianVmariano** [comment]
Oh, and it has nothing to do with the override business.  The child of attachable needs to be the centered, original shape.  Almost all the recently written stuff works like rect() does where the function makes the shape and the module just calls the function.  (Some earlier code is more redundant because Revar didn't come up with the idea of making function forms for everything until later and so some of the functions were written after the modules.)   But when the module calls the function, you **never** want to pass the anchor, spin or orient because you want attachable() to manage that stuff in the module.  That's necessary for the attachment features to work properly.

**Comment by amatulic** [comment]
One more thing. It doesn't seem to matter if I set `extent` to true or false in `attachable()` and `reorient()`, the anchors always look the same.

`squircle(0.5, [164,48], anchor=BOTTOM+LEFT, atype="perim", spin=20) show_anchors();`
image
Will that be a problem?

Before approval, do you prefer I add in the negative squareness support? Or leave that for later?

**Comment by adrianVmariano** [comment]
The extent option doesn't do anything if you give size.  It's only for region based anchors, not for size-based rectangular anchors.  

However, in looking at your anchors shown above it does seem like the corners are not right.  They don't look like they are on the natural corner.  Did you get those by making a square and then stretching it?   Because that won't result in the actual corners.   Note that they are both oddly positioned and not perpendicular to the curve?  

You can worry about negative squareness later.   Note that I'm not insisting on it, just noting it as a possibility so that if we skip it that is a **choice** and not an oversight.  Let's get it processed through all the checks.  They are quite slow (like half an hour) because it reprocessed the entire wiki.   I don't recall if I have to re-approve it after you make changes, or if once it's been approved, it stays approved.

**Comment by amatulic** [comment]
The oblong squircle is made by scaling the x and y values according to the size vector. Whatever vertex was at the corner with equal x and y sizes no longer has the maximum radial position when stretched.

Here's the same image with the symmetric squircle overlaid. The arrows are still attached to the enongated one, but you can see they're normal to the symmetric squircle. I expected this to happen, but I was not sure if a user would prefer to anchor at the original squircle perimeter point or a different one from stretching.
image

**Comment by amatulic** [comment]
Another question: If it is preferable for the anchor to be at the maximum radial position on the perimeter, would that be the actual vertex in the path with the maximum radius, or the mathematical maximum radius? I could do either one; the first is easiest, the latter would require a recursive iteration to find the mathematical maximum-radius [x,y] point on the stretched squircle and then repeat at point around all four anchors.

**Comment by amatulic** [comment]
Hmm, I was wrong, the desired anchor isn't at the maximum radial point, it's where the slope of the curve is 45° plus the spin angle.

**Comment by adrianVmariano** [comment]
Note also that you don't need to use recursion to find the max.  

But yeah, the point of interest is the one at 45 degrees.  I think maybe you can also find it by mapping backward from the rectangle corner to the unit square---not not certain as I haven't tried to work out details for that.  

We have done both things in BOSL2 in different places:  sometimes anchors are on the idealized curve regardless of sampling (e.g. spheres, cylinders) and sometimes anchors are on the realized path point.   One major challenge with some of the weird anchors is that I don't actually know what they are good for.  So why would a user want an anchor on the rounded corner of a squircle rectangle?   How will they use it?  The argument for ideal is that then the anchors don't shift around if you fiddle with $fn.  The argument for actually on the path is that then the anchor is, well, actually on the path, so the path **actually** goes where it was anchored or an attached object really touches there.   

It looks like other shapes in shapes2d.scad that do this (rect and trapezoid) both use line intersection calculations on the path to find the **actual** intersection point.  (I think that works because the centerpoint of the rounding is known.)   So consistency suggests that actually on the path is best.  But if you feel like you have a strong case for idealized, you can do it that way, since as I've said, things are not consistent and it's done that way sometimes.  It would be good to document the behavior, whatever it is.  (I'm not sure how well documented existing stuff is.)  

Note that when you do your calculations, the spin angle is not involved.  It gets applied afterwords by the attachment engine.  You are working on the shape in its native position.

**Comment by adrianVmariano** [comment]
I'm a bit confused because your description sounds like a mixture of idealized and "on the curve".  If you're doing idealized you should be able to just compute it from the functional form.  If you're doing on the curve, it should actually be on the curve for all $fn, and probably the normals should be normal to the facet it's sitting on.  

BTW, I realized I totally botched the computation of the polar form for FG that I used for previous graphs by losing a factor of r^2 which made things oh so much simpler.  My basic point still stands about the shapes---just the graphs are wrong.  Correct graphs (not from polar coordinate equation):
image

**Comment by amatulic** [comment]
Oh, I see. I could use the derivative of the cartesian form of the FG and superellipse in one quadrant to find where it's exactly 45° and then use that angle, instead of interpolating from the existing polygon.

I haven't looked into the negative squareness yet, but your amendment above shows what I suspected, that the FG version doesn't result in a diamond.

**Comment by adrianVmariano** [comment]
And yes, use the derivative of the cartesian form (presumably the simpler approach) of the scaled shape and set the slope equal to -1 and solve to find corner location.  (The angle will of course be 45 by construction.)  

Yeah, that's right, you never get a perfect diamond with corners.  (You didn't in my botched computation either....)  You get a rounded diamond shape when (unsquared) s=-4, which I included in the graph.  That's the curve in the family which is tangent to the diamond.

**Comment by adrianVmariano** [comment]
Hmmm.  I have a suspicion the solution will be impossible in closed form super superellipse and maybe also for FG. Haven't tried to really do the computation so not sure, but might need to use root_find() to pursue this approach.

**Comment by amatulic** [comment]
Both have a closed-form derivative. I used Wolfram Alpha to solve them. The cartesian FG equation is turned 45° from the polar form so that isn't as useful to figure out the corner position of a stretched version, but the derivative of the polar FG with respect to theta is well behaved. The problem I still have to figure out is how to stretch the polar form.

**Comment by amatulic** [comment]
Eh... both of us dropped a minus sign. In your picture, the s^2 x^2 y^2 term should be negated.

**Comment by amatulic** [comment]
And you're right, while there is a closed form derivative of both FG and superellipse, solving for where the slope=-1 would require a numerical approach. Which brings me back to my implemented solution of interpolating the vertex angle normals to find the anchor angle corresponding to a 45° slope. I'm thinking if I'm going to approximate it anyway, that's probably close enough, especially if `$fn` is at least 40 or so.

And then there's the point you made about why anyone would even want to do anchor the corner of the squircle. All my uses of a squircle have been relative to its center or the squircle's bounding box.

**Comment by adrianVmariano** [comment]
Consider this curve.  It seems to me that the 45 deg point on the "corner" is not a very good representative of the corner of the shape.  I also took a look at ellipse() which produces kinda garbage corner anchors (based on 45 deg line intersection point.)  

image

I therefore think that the following reasonably straight forward approach might be good:  Find the intersection between the curve and the ray from the origin to the corner of the bounding rectangle.  Use the actual intersection with the curve and the angle of the facet that you find.   

So something along the lines of 
```
corner=[for(seg=pair(pts)) let(isect=line_intersection(seg, [[0,0],[a,b]],SEGMENT,RAY)) if (is_def(isect) [isect,-line_normal(seg)]
//  corner might have 2 solutions if the solution was a vertex, shared by both segments, so in that case take the average normal
```

**Comment by amatulic** [comment]
That's what I was doing originally. That intersection is always going to be the same point in the polygon path, whether it's stretched or not. That was this picture above:
image

My concern there was the anchor arrows weren't coming out of the figure perpendicular to the perimeter, and you responded that this works for region-based anchors, not size-based anchors.

**Comment by adrianVmariano** [comment]
I somehow failed to recognize that these anchors are the same as the 45 deg anchors on the unit shape.   I still think they look sort of like they're in the wrong place....but it seems like everything else we've considered also is somehow wrong.  So I'm suggesting just use these anchor positions, but adjust the direction so they are perpendicular to the perimeter.   Or if you want to do the 45 deg angle point on the curve that seems OK as well.  (Maybe it looks better?  I don't know.  It seems more difficult, and I'm trying to prioritize ease.)  

Using root_find:
```
f = function (z) z*sin(z)-4;
rt=root_find(f, 0,25);

amatulic:
I just noticed that the superellipse generated by supershape() has an anchor at 45° for a stretched squircle. For consistency, perhaps we should do that here too.

This has been an interesting mathematical exercise, but it's been a lot of investigation for a feature that is unlikely to be used. Let me see what I can do with root_find().

amatulic:
OK, got it with root_find() now. The image I posted for the superellipse was missing the 'b' term from the derivative. I'll tidy up my comments and commit another change to this PR.

adrianVmariano:
When you have a weird shape, it uses path based (or region based) anchors where it just intersects with a line at 45 degrees.  That doesn't mean that this choice is the best thing to do, it's rather the least stupid thing we could come up with when faced with a totally arbitrary shape.    In the context of supershape so much can happen that reasonable anchors are impossible.  With squircle that's not true.   There is no reason at all to duplicate the supershape anchors here.  In fact, part of the value added in having a separate function is having smarter anchors.   Someone **might** want to put some kind of object on the "corners" of the shape when building up some kind of object, so I wouldn't say never.  But I definitely understand where you are coming from.  I feel like I have spent hours trying to fix anchors for edges and corners of 3d objects that seemed unlikely to ever be used for anything.  Though I did finally recognize an important use case for edge anchors.  (Corner anchors, however, remain a mystery:  I don't know of any application that use the direction of corner anchors on 3d objects, or the position of corner anchors on curved corners.)

amatulic:
OK, do you prefer going back to the original anchors where the diagonals of the rectangle intersect the squircle perimeter? It isn't a problem either way. I just have no idea what's best, and no idea if anyone would even want to anchor at a corner.

adrianVmariano:
Did you actually implement 45 deg anchors?  If yes, then let's see what they look like for extreme aspect ratios so we can see if one looks somehow better.  If not, do less work and use the method already implemented.

amatulic:
Yes, I did actually implement the 45° anchors using the derivative functions for both types of squircles, using find_root(). It's in the latest commit I pushed. They look exactly like the the anchors you get from supershape() when you make a squircle with it.

It's very easy to go back to the box diagonal intersections, however, and that code is a bit simpler too.

adrianVmariano:
Oh, supershape must be using "hull" style anchors.   I wasn't thinking about that...those anchors would of course be at 45 degrees.   I had my mind stuck on the "intersection" style anchors like ellipse is using.  You can tell that I really use these anchors, eh?   So maybe I should reverse my previous statement about using the supershape anchors.  That might actually be the right thing---and easiest thing---to do.  

To do that you would invoke attachable with extent=true and with a path given and no size argument given.   It computes the anchors from the path by looking for the most distance point that intersects a line perpendicular to the anchor direction.  

Probably ellipse() should be changed to use these anchors instead.  I see that actually it's using the "circle" anchors which kinda make no sense for skinny ellipses.   Actually if you want to see an example of a math problem that got out of hand ellipse() is such a case.  I thought it would be nice to produce ellipses with all the edges the same length.   Turned out to require an iterative algorithm.   And I never did find a robust solution for the case of a circumscribing ellipse.  

Ellipse anchors:

image

Note how they are the points of intersection of a 45 deg angle line, which doesn't make much sense in this case.   I wonder if ellipse was written before hull anchoring.

adrianVmariano:
I guess one problem with choosing "hull" anchors for ellipse is that then the anchors aren't the same as the circle anchors for the case of circles---because circle anchors are ideal and the "hull" anchors are on the shape.   But this obviously has nothing to do with squircle.

amatulic:
>you would invoke attachable with extent=true and with a path given and no size argument given

There seems to be more to it than that. I changed the `attachable()` call, specifying the path and extent=true, made the override path `undef`, and I got an error message "is_region(region) failed".

If this anchor type uses the path, the anchors would not be on the mathematical perimeter, but on one of the segments in the path, right?

adrianVmariano:
Yes, if you use the extent=true "hull" type anchoring the anchors will be on the path, not on the mathematical perimeter (which is not available to the code so obviously impossible).  

The yellow squircle seems to have a really large facet right around where the anchor is being placed while the rest of the visible perimeter looks smooth.  Is that right?  

So is this ready to be approved?

amatulic:
That yellow squircle was made with $fn=40. That's only 10 segments per quadrant, and the segments are made smaller at the corners. So the segment there at the anchor is actually smaller than it normally would be at that $fn value. You can see this with 'show edges' enabled:

image

I observed that if $fn is divisible by 8, the anchors line up with the superellipse anchors. At least for this case. I may as well change the default $fn anyway.

Not yet ready -- I have to remove all the stuff about anchor overrides, then I'll do a bit more testing, and push another commit.

Hard to believe I spent a couple of days on this only to delete it all, but at least I learned a lot, and I appreciate that.

adrianVmariano:
I kinda feel a little guilty about having lead you on a run-around on the anchoring business here.  Sorry about that.  I'll admit that it seems pretty routine that adding some new object is 10% coding the object and 90% coding either the anchors or the parameter error checking and parameter processing, or some other indirect thing relating to the object rather than the object itself.   Someone recently asked for ring() which was in that category, where making a ring is trivial, but doing all the argument checking for all the cases...ug.  

Wouldn't the alignment to superellipse anchors depend on the step size used for the superellipse?  If you configure them to match then they should line up...otherwise no.  

It seems like there's a sharp corner in the example above right where the anchor connects.   It's a little easier to see what's happening if you view 2d stuff directly from above (e.g. click on the z+ view).   But I'm sure it gets better with higher $fn, so I doubt there's anything worrisome there.

amatulic:
Yes, that sharper corner is an artifact of low $fn. I agree the alignment with supershape likely depends on step size, but in any case a high value of $fn makes them align quite well. Hmm, I should have it change dynamically like it does with circle(), with more steps for larger shapes.

Don't feel guilty, I appreciate the knowledge, and it was valuable for me to learn about function literals, which in all the years I've been working with OpenSCAD, I never noticed before but often wished for such a capability.

adrianVmariano:
I'm not sure what you mean about dynamic change "like circle(), with more steps for larger shapes".  That is if you use $fa and $fs.  With $fn, circle always gives you the specified number of facets.

adrianVmariano:
BTW, function literals were only added in 2021.   But hang out with us library authors and we'll teach you all the OpenSCAD tricks.  :)

amatulic:
I mean if you don't specify $fn, a circle changes the number of segments depending on the size of the circle (at least for small circles), because it uses $fa=12 together with $fs=2 by default. I thought I might increase the resolution of the squircle if $fn isn't set, based on the radius of the circle when squareness=0.

adrianVmariano:
Your original stated plan was to respect only $fn.  It seems reasonable to respect $fa and $fs based on the circle corresponding to the specified size, though it's not entirely obvious how to handle that for ellipses.   I looked at ellipse().  It treats the largest semi-axis as the effective radius for calculating the number of segments of the curve.  

There is infrastructure in place to do this in BOSL2 with the segs() function.

amatulic:
I am hopelessly confused now. I'm obviously missing something. Three times now, I've restored the original code and change it a bit at a time, and at some point it stops respecting the atype="perim", and I don't know what I did. Here's the module and function. Does anything jump out at you?

````
module squircle(squareness=0.5, size=[1,1], style="fg", atype="box", anchor=CENTER, spin=0) {
    check = assert(squareness >= 0 && squareness <= 1);
    anchorchk = assert(in_list(atype, ["box", "perim"]));
    size = is_num(size) ? [size,size] : point2d(size);
    assert(all_positive(size), "All components of size must be positive.");
    path = squircle(squareness, size, style, atype); // pass without anchor
    if (atype == "box") {
        attachable(anchor, spin, two_d=true, size=size, extent=false) {
            polygon(path);
            children();
        }
    } else { // atype=="perim"
        attachable(anchor, spin, two_d=true, size=size, extent=true, path=path) {
            polygon(path);
            children();
        }
    }
}


function squircle(squareness=0.5, size=[1,1], style="fg", atype="box", anchor=CENTER, spin=0) =
    assert(squareness >= 0 && squareness <= 1)
    assert(is_num(size) || is_vector(size,2))
    assert(in_list(atype, ["box", "perim"]))
    let(
        size = is_num(size) ? [size,size] : point2d(size),
        path = style == "fg" ? _squircle_fg(squareness, size)
            : style == "superellipse" ? _squircle_se(squareness, size)
            : assert(false, "Style must be \"fg\" or \"superellipse\""),
    ) reorient(anchor, spin, two_d=true, size=size, p=path, extent=atype=="perim");
````

adrianVmariano:
You are giving a size argument to attachable() in the "perim" branch, that case I specifically warned about earlier where the code decides it's going to "do something" instead of failing with an assert when giving conflicting inputs.  (Are you convinced yet that this "do something" strategy is a terrible idea?)  

Unrelated: I would prefer not to have a default for size.  Revar is more a fan of defaults like that.  I tend to think they're random and arbitrary and the user should say what they mean.   But rect() does have a default so you can choose to follow rect if you prefer.

amatulic:
Ah, that was it, the size. I wonder why I didn't see it. That fixed the module, but calling the function to use with polygon() is still using atype="box".

As for defaults, I did that because OpenSCAD uses a default size for things like square() and circle(). I can remove it, however.

adrianVmariano:
I notice you didn't say whether your perspective on asserting for conflicting inputs has changed.  

I'm not sure what you mean about the function case.  If you remove the size parameter to reorient in the atype=="perim" case is something wrong?  And if so, what exactly?  Are you saying the shape comes out centered regardless of the anchor value?   (Of course you need the size parameter for the "box" case.)  

I would maybe not pass atype when invoking the function form from the module.  It's presumably best to have the simpler anchor case run, though I'm not sure it makes any difference since you just get the center anchor.  

I think you can set extent to true unconditionally because it's only used when no size is given.   But in the module form I would not set extent in the "box" branch just because it's confusing to pass a parameter that's not used.

amatulic:
"Perspective for asserting for conflicting inputs". I am asserting for conflicting inputs, rather than falling back to defaults. Regardless of my own way of coding, I want to do whatever is considered best practice by this project.

Yellow is the module, outline is the polygon, both using the same anchor and atype.
image

amatulic:
As soon as I submit my last comment, I figure out the problem.

It works with supershape. In reorient(), the supershape function sets two parameters, p and path, to path. It also omits the size. My call to reorient() was left over from the older override anchor code, setting size and p but not path. I did it like supershape, and it all works now.

adrianVmariano:
Are you doing something like `size=atype=="perim"?undef:size` to pass the size to reorient as undef in the perim case and the actual size in the box case?  

I had forgotten you had to pass the path twice.  I think it's one time for purposes of anchoring and one time to just be the path that gets processed, so you don't have to use the path as the anchoring method.

amatulic:
I ended up doing it like this:
````
reorient(anchor, spin, two_d=true, size=atype=="box"?size:undef, path=_module_call?undef:path, p=path, extent=true);
````
I had to add a parameter `_module_call` to the function because if you set both p and path when calling it from the module, the shape is positioned incorrectly. `size` depends on atype, and `extent=true` works in all cases.

amatulic:
All right. I have pushed the latest changes. Have a look. No hurry.
Now I can continue testing vnf_tri_array() and make new examples.

revarbat:
Of note, `circle()` and `square()` have default sizes in raw OpenSCAD, so BOSL2's overrides for those *should* match that behavior. Whether `rect()` or other BOSL2 originals should have size defaults is a separate argument.

revarbat:
I used to be of the opinion that args should have "reasonable" defaults. But that gets nonsensical sometimes, and my idea of reasonable is not necessarily the most usable or sensible for all situations. I'm starting to turn to the dark side of not giving defaults, but giving examples that seem sensible to me.

amatulic:
I can't say I understand those test results. There's a syntax error being reported in shapes2d.scad, referencing a variable name that was used in an older commit but no longer exists, making me think the tests weren't run on the most recent commit.

adrianVmariano:
You have a trailing comma on line 2054 in the let statement.  That is not allowed in the stable version 2021.01.  Since that's the official stable version, everything needs to work there, and the trailing comma thing is one of the big irritations, most definitely.  

BTW, you shouldn't need that _module_version or whatever it was flag.  Just make sure to pass atype="box" when invoking from the module.  Don't pass the actual user-specified atype.

amatulic:
Hopfull that last commit passes the tests. I was using "Examples" instead of "Example" for a multi-line code example.

amatulic:
Sigh. Another error in the documentation comments. A bad internal link. OK, I pushed another fix.

PR #1520, opened 2024-12-07
Update link to linear_sweep

Skinner927:


attachments.scad
```diff
@@ -1155,7 +1155,7 @@ module tag_this(tag)
 //   - `polygon()`
 //   - `projection()`
 //   - `polyhedron()`  (or use [`vnf_polyhedron()`](vnf.scad#vnf_polyhedron))
-//   - `linear_extrude()`  (or use [`linear_sweep()`](regions.scad#linear_sweep))
+//   - `linear_extrude()`  (or use [`linear_sweep()`](skin.scad#functionmodule-linear_sweep))
```
revarbat, line 1158: I believe the easier way to make that link would be to use `{{linear_sweep()}}` which will automatically track is that function is moved between files.

adrianVmariano:
I added this to my PR using the new link syntax.

PR #1521, opened 2024-12-11
Replaced vnf_tri_array(), added example to vnf_from_polygons()

amatulic:
Changes:
* Replaced vnf_tri_array() with version that does not restrict row length or row wrapping, added `row_wrap`, `cap1`, `cap2`, and `caps` parameters for consistency with vnf_vertex_array()
* Edited documentation for vnf_tri_array
* Modified examples and added more examples
* Added example to vnf_from_polygons(), which didn't have any
* Corrected some typos in comments

I tested this against all of the examples in [rounded_prism](https://github.com/BelfrySCAD/BOSL2/wiki/rounding.scad#functionmodule-rounded_prism) too.

Here's the final example I made with vnf_tri_array(). I think it illustrates the usage of differing number of row points quite well, to make an object with roughly same-size facets.

image

### DIFF
--- a/vnf.scad
+++ b/vnf.scad
@@ -318,8 +318,8 @@ function vnf_vertex_array(
 // Example(3D,NoAxes,Edges): Model of a cymbal with roughly same-size facets, using a different number of points for each concentric ring of vertices.
 //   include <BOSL2/beziers.scad>
 //   bez = [
-//       [[0,22], [35,22], [30,0], [80,14], [102,0]], //top
-//       [[99,-1], [79,13], [29,-1], [34,21], [-1,21]] // bottom
+//      [[0,24], [35,24], [30,0], [80,15], [102,0]], //top
+//      [[99,-1], [79,14], [29,-1], [34,23], [-1,23]] // bottom
 //   ];
 //   points = [
 //       for(b=bez)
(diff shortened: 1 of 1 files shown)


adrianVmariano:
Is that example flat or is it like a sine wave?

amatulic:
That's just weird. The checks failed due to one critical line that got removed somehow just before my last save. It probably happened while I was cleaning up comments.

It's supposed to be a percussion cymbal, with a raised bell in the middle and a bow-shaped curve down to the edge. I could exaggerate the height more. The profile isn't a sinewave, it's a quartic bezier curve.

amatulic:
OK, I retested. I don't know what happened there, maybe I inadvertently hit "undo" before I saved it. I retested a few things, and what got broken is now working. Please re-check at your convenience.

adrianVmariano:
I didn't mean sine wave specifically.   I suppose 3 dimensionality isn't important to the example.  I'll leave it up to you if you want to fiddle with it.  You might also be able to make it look more 3d by changing the viewpoint, by specifying VPR, VPT and VPD in the example line.

amatulic:
Oh, good, the checks passed.

I did slightly increase the height of the cymbal model before pushing that last change. Not much difference. I'd like to see how it looks in the documentation. There may be some other examples I need to tweak. I did adjust the camera angle for the example I added to vnf_from_polygons because the shading made too many facets the same color, although I guess I could have turned on edges to distinguish them.

amatulic:
OK, since you brought it up, I'm finding myself more bothered by the view angle of the cymbal. I changed it to a shallower view by 10° , and also found a redundant comment block, which I removed.

Issue #1559, opened 2025-02-03
Examples for possible inclusion in isosurface.scad


RAMilewski:
Note that these use a local copy of isosurface.scad so line 2 must be changed before inclusion.

RAMilewski:
<img width="1043" alt="Image" src="https://github.com/user-attachments/assets/29199e63-21a1-4050-8a56-90d88f202239" />

'''
include <BOSL2/std.scad>
include <isosurface.scad>

stats = false; // [true,false]

bounding_box = [[-60,-50,-5],[35,50,21]];
funcs = [
    move([-20,0,0]) * scale([25,4,4]),   mb_sphere(1),      // fuselage
    move([30,0,5])  * scale([4,0.5,8]),  mb_sphere(1),      // vertical stabilizer
    move([30,0,0])  * scale([4,15,0.5]), mb_sphere(1),      // horizontal stabilizer
    move([-20,0,0]) * scale([6,45,0.5]), mb_sphere(1),      // wing
];  
isovalue = 1;
voxel_size = 1;

metaballs(funcs = funcs, isovalue=isovalue, bounding_box = bounding_box, voxel_size = voxel_size, show_stats = stats);

'''

amatulic:
Nice! Shrinking the bounding box just a little bit, to `[[-55,-50,-5],[35,50,17]]`, improved the execution time by 1.5 seconds.

RAMilewski:
Ha... I'm also a pilot (SEL and LTA with about 25 solo hours in sailplanes but no sailplane rating).     I kept the wing forward to keep the aft fuselage from getting too fat.   That version was my first rough cut.   If it can get the time I'll maybe add a cockpit canopy and tip-tanks.   Not sure if I can turn it into an airliner with underwing engines.

...any idea how to do a NACA airfoil in metaballs?  :)

amatulic:
Cool. My ratings are single-engine land and glider aero-tow-only, and most "recently" I got an ultralight certification (before the recreational pilot rating existed). All this happened in Texas in the 1970s and 1980s when costs were reasonalble. I haven't flown in years, however. My last flight was to take my fiancee (now my wife) up for a spin in Virginia where I lived at the time. I live in California now, and it's too expensive to fly around here.

I was thinking about an airfoil too. NACA is a rather precise formula, so I don't think that's possible with metaballs, but you may get something close by using negative metaballs to subtract the lower camber.

RAMilewski:
[...]

PR #1566, opened 2025-02-16
Doc fixes, code changes, bugfixes, new examples

amatulic:
Captured suggestions from chat. Still to do is add a new example of a custom function and renumber the example references in the text.

@adrianVmariano , @RAMilewski :
Please add any new suggestions to this PR.

### DIFF
--- a/isosurface.scad
+++ b/isosurface.scad
@@ -739,7 +739,7 @@ function _isosurface_cubes(voxsize, bbox, fieldarray, fieldfunc, isovalmin, isov
         for(x=[v.x:voxsize:b1.x]) [
             for(y=[v.y:voxsize:b1.y]) [
                 for(z=[v.z:voxsize:b1.z])
-                    fieldfunc(x,y,z)
+                    fieldfunc([x,y,z])
             ]
         ]
     ],
@@ -1441,10 +1441,9 @@ function mb_octahedron(r, cutoff=INF, influence=1, negative=false, d) =
 //   number of voxels below 10,000 for preview, and adjust the voxel size smaller for final
 //   rendering.  A bounding box that is larger than your isosurface wastes time computing function
 //   values that are not needed. If the metaballs fit completely within the bounding box, you can
-//   call {{vnf__bounds()}} on the VNF structure returned from the `metaballs()` function to get an
-//   idea of a the optimal bounding box to use.  You may be able to decrease run time, or keep the
-//   same run time but increase the resolution. You can also set the parameter `show_stats=true` to
-//   get the bounds of the voxels containing the generated surfaces.
+//   set the parameter `show_stats=true` to get the bounds of the voxels containing the generated
+//   surfacesto get an idea of a the optimal bounding box to use.  You may be able to decrease run
+//   time, or keep the same run time but increase the resolution. .
 //   .
 //   The point list in the returned VNF structure contains many duplicated points. This is not a
 //   problem for rendering the shape, but if you want to eliminate these, you can pass
--- a/isosurface.scad
+++ b/isosurface.scad
@@ -831,7 +831,7 @@ function _isosurface_triangles(cubelist, voxsize, isovalmin, isovalmax, tritable
 
 /*
 /// Generate triangles for the special case of voxel faces clipped by the bounding box
-// (more efficient than _bbfacevertices below but doesn't work with isovalue ranges)
+/// (more efficient than _bbfacevertices below but doesn't work with isovalue ranges)
 function _clipfacevertices(vcube, f, bbface, isovalmax, isovalmin) =
     let(
         vi = _MCFaceVertexIndices[bbface], // four voxel face vertex indices
@@ -1928,7 +1928,7 @@ function _mb_unwind_list(list, parent_trans=[IDENT]) =
 //   f = The isosurface function or array.
 //   isovalue = A 2-vector giving an isovalue range. For an unbounded range, use `[-INF, max_isovalue]` or `[min_isovalue, INF]`.
 //   voxel_size = size of the voxel that is used to sample the bounding box volume. This can be "auto", a scalar size for a cubical voxel, or a 3-vector if you want non-cubical voxels. For "auto", the voxel size is set so that approximately `auto_voxels` (default 8000) quantity of voxels fit inside the bounding box. If you set `fixed_bounds=true`, then bounding box is held fixed in size, and the voxel size is adjusted as needed so that whole voxels fit inside the bounding box.
-//  bounding_box = A designation of volume in which to perform computations, expressed as a scalar size of a cube centered on the origin, or a pair of 3D points `[[xmin,ymin,zmin], [xmax,ymax,zmax]]` specifying the minimum and maximum box corner coordinates. By default, the actual bounding box is enlarged if necessary to fit whole voxels, and centered around your requested box. Set `fixed_bounds=true` to hold the box size fixed, in which case the voxel changes size instead. When `f` is an array of values, `bounding_box` is already implied by the array size combined with `voxel_size`, in which case this implied bounding box is centered around the origin.
+//   bounding_box = A designation of volume in which to perform computations, expressed as a scalar size of a cube centered on the origin, or a pair of 3D points `[[xmin,ymin,zmin], [xmax,ymax,zmax]]` specifying the minimum and maximum box corner coordinates. By default, the actual bounding box is enlarged if necessary to fit whole voxels, and centered around your requested box. Set `fixed_bounds=true` to hold the box size fixed, in which case the voxel changes size instead. When `f` is an array of values, `bounding_box` is already implied by the array size combined with `voxel_size`, in which case this implied bounding box is centered around the origin.
 //   ---
 //   closed = When true, close the surface if it intersects the bounding box by adding a closing face. When false, do not add a closing face and instead produce a non-manfold VNF that has holes.  Default: true
 //   reverse = When true, reverses the orientation of the VNF faces. Default: false
@@ -2149,7 +2149,7 @@ function isosurface(f, isovalue, voxel_size, bounding_box, reverse=false, closed
         dum2 = show_stats ? _showstats_isosurface(voxsize, bbox, isovalue, cubes, trianglepoints, faces) : 0
 ) [trianglepoints, faces];
 
-// internal function: get "auto" voxel size given a desired number of voxels in a bounding box
+/// internal function: get "auto" voxel size given a desired number of voxels in a bounding box
 function _getautovoxsize(bbox, numvoxels) =
     let(
         bbsiz = bbox[1]-bbox[0],
@@ -2157,7 +2157,7 @@ function _getautovoxsize(bbox, numvoxels) =
         voxvol = bbvol/numvoxels
     ) voxvol^(1/3);
 
-// internal function: get voxel size, adjusted if necessary to fit bounding box
+/// internal function: get voxel size, adjusted if necessary to fit bounding box
 function _getvoxsize(voxel_size, bounding_box, fixed_bounds) =
     let(voxsize0 = is_num(voxel_size) ? [voxel_size, voxel_size, voxel_size] : voxel_size)
         fixed_bounds ? 
--- a/isosurface.scad
+++ b/isosurface.scad
@@ -77,49 +77,7 @@ Vertex and edge layout (heavier = and # indicate closer to viewer):
    0 +==========+ 4          +=====8=====+
 
 z changes fastest, then y, then x.
-
------------------------------------------------------------
-Addition by Alex:
-Vertex and face layout for triangulating one voxel face that corrsesponds to a side of the box bounding all voxels.
-
-                    4(back)
-               3 +----------+ 7
-                /:  5(top) /|
-               / :        / |
-            1 +==========+5 |    <-- 3(side)
-0(side) -->   # 2+ - - - # -+ 6
-              # /        # /
-              #/  2(bot) #/
-            0 +----------+ 4
-                1(front)
 */
-
-/// four indices for each face of the cube, counterclockwise looking from inside out
-_MCFaceVertexIndices = [
-    [],
-    [0,2,3,1], // left, x=0 plane
-    [0,1,5,4], // front, y=0 plane
-    [0,4,6,2], // bottom, z=0 plane
-    [4,5,7,6], // right, x=voxsize plane
-    [2,6,7,3], // back, y=voxsize plane
-    [1,3,7,5], // top, z=voxsize plane
-];
-
-/// return an array of face indices in _MCFaceVertexIndices if the voxel at coordinate v0 corresponds to the bounding box.
-function _bbox_faces(v0, voxsize, bbox) = let(
-    a = v0-bbox[0],
-    bb1 = bbox[1] - [voxsize,voxsize,voxsize],
-    b = v0-bb1
-) [
-    if(a[0]==0) 1,
-    if(a[1]==0) 2,
-    if(a[2]==0) 3,
-    if(b[0]>=0) 4,
-    if(b[1]>=0) 5,
-    if(b[2]>=0) 6
-];
-/// End of bounding-box face-clipping stuff. Back to the marching cubes triangulation....
-
 
 /// Pair of vertex indices for each edge on the voxel
 _MCEdgeVertexIndices = [
@@ -671,6 +629,95 @@ function _cubeindex(f, isoval) =
     (f[6] > isoval ? 64 : 0) +
     (f[7] > isoval ? 128 : 0);
 
+/*
+-----------------------------------------------------------
+Bounding box clipping support:
+
+Vertex and face layout for triangulating one voxel face that corrsesponds to a side of the box bounding all voxels.
+
+                    4(back)
+               3 +----------+ 7
+                /:  5(top) /|
+               / :        / |
+            1 +==========+5 |    <-- 3(side)
+0(side) -->   # 2+ - - - # -+ 6
+              # /        # /
+              #/  2(bot) #/
+            0 +----------+ 4
+                1(front)
+
+The clip face uses different indexing. After vertex coordinates and function values are assigned to each corner from the original voxel based on _MCFaceVertexIndices below, this is the clip face diagram:
+
+(1)           (2)
+   +----1----+
+   |         |
+   0         2
+   |         |
+   +----3----+
+(0)           (3)
+*/
+
+/// four indices for each face of the cube, counterclockwise looking from inside out
+_MCFaceVertexIndices = [
+    [],
+    [0,2,3,1], // left, x=0 plane
+    [0,1,5,4], // front, y=0 plane
+    [0,4,6,2], // bottom, z=0 plane
+    [4,5,7,6], // right, x=voxsize plane
+    [2,6,7,3], // back, y=voxsize plane
+    [1,3,7,5], // top, z=voxsize plane
+];
+
+/// Pair of vertex indices for each edge on the clip face (using clip face indexing)
+_MCClipEdgeVertexIndices = [
+    [0,1], [1,2], [2,3], [3,0]
+];
+
+/// For each of the 16 configurations of a clip face, define a list of triangles, specified as pairs of corner ID and edge ID arrays, with a total of 3 points in each pair. Each pair has the form [corner],[edge1,edge2] or [corner1,corner2],[edge].
+/// In keeping with the convention for triangulating an isosurface through a voxel, analogous to the case in which two surfaces separate two diagonally opposite greater-than-isovalue corners of one face, in 2D contour terms it is assumed there is a valley separating two diagonally-opposite high corners, not a ridge connecting them. The two triangulation cases for opposing corners are set up accordingly.
+_MCClipTriangleTable = [
+    [],                     // 0 - 0000 - ignored
+    [[0],[0,3]],            // 1 - 0001
+    [[1],[1,0]],            // 2 - 0010
+    [[0,1],[1], [0],[1,3]], // 3 - 0011
+    [[2],[2,1]],            // 4 - 0100
+    [[0],[0,3], [2],[2,1]], // 5 - 0101 - opposing corners
+    [[1,2],[0], [2],[2,0]], // 6 - 0110
+    [[0,1],[3], [1],[2,3], [1,2],[2]], // 7 - 0111
+    [[3],[3,2]],                       // 8 - 1000
+    [[3,0],[0], [3],[0,2]],            // 9 - 1001
+    [[1],[1,0], [3],[3,2]],            //10 - 1010 - opposing corners
+    [[0,1],[1], [0],[1,2], [3,0],[2]], //11 - 1011
+    [[2,3],[3], [2],[3,1]],            //12 - 1100
+    [[3,0],[0], [3],[0,1], [2,3],[1]], //13 - 1101
+    [[2,3],[3], [2],[3,0], [1,2],[0]], //14 - 1110
+    [[0,1,2],[], [0,2,3],[]],          //15 - 1111
+];
+
+/// _clipfacindex() - private function, called by _clipfacevertices()
+/// Return the index ID of a voxel face depending on the field strength at each corner exceeding isoval.
+function _clipfacindex(f, isoval) =
+    (f[0] > isoval ? 1 : 0) +
+    (f[1] > isoval ? 2 : 0) +
+    (f[2] > isoval ? 4 : 0) +
+    (f[3] > isoval ? 8 : 0);
+
+/// return an array of face indices in _MCFaceVertexIndices if the voxel at coordinate v0 corresponds to the bounding box.
+function _bbox_faces(v0, voxsize, bbox) = let(
+    a = v0-bbox[0],
+    bb1 = bbox[1] - [voxsize,voxsize,voxsize],
+    b = v0-bb1
+) [
+    if(a[0]==0) 1,
+    if(a[1]==0) 2,
+    if(a[2]==0) 3,
+    if(b[0]>=0) 4,
+    if(b[1]>=0) 5,
+    if(b[2]>=0) 6
+];
+/// End of bounding-box face-clipping stuff
+/// -----------------------------------------------------------
+
 
 /// isosurface_cubes() - private function, called by isosurface()
 /// This implements a marching cubes algorithm, sacrificing some memory in favor of speed.
@@ -753,11 +800,7 @@ function _isosurface_triangles(cubelist, cubesize, isovalmin, isovalmax, tritabl
                 v, v+[0,0,cubesize], v+[0,cubesize,0], v+[0,cubesize,cubesize],
                 v+[cubesize,0,0], v+[cubesize,0,cubesize],
                 v+[cubesize,cubesize,0], v+[cubesize,cubesize,cubesize]
-            ],
-            outfacevertices = flatten([
-                for(bf = bbfaces)
-                    _bbfacevertices(vcube, f, bf, isovalmax, isovalmin)
-            ])
+            ]
         )
         each [
             if(len(tritablemin[cbidxmin])>0) for(ei=tritablemin[cbidxmin]) // min surface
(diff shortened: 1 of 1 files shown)


adrianVmariano:
Ok.  Reading over the code and also a few more comments from others:

1. Negative metaballs: delete bit about "are below the isovalue" (we haven't mentioned isovalue).  Just say "negative metaballs are never directly visible
2. Add something about how influence behaves in a reversed (?) or unexpected way at the center of negative metaballs
3. Text about isovalue parameter sort of makes it seem like it's a ball parameter instead of a global parameter when the paragraph starts.  It is feels like an injection between a paragraph about model design and then the following paragraph on the same topic.    So I think move the isovalue paragraph down to be the last paragraph before the built-in functions list, maybe?  Or some more clear transition, like "Another global metaball parameter is..."
4. I wonder if the user function section should say: "where $d$ is a distance measured from the center or core of the metaball."  That better captures what happens for capsule, disk and cyl.   Then the later sentence seems somewhat unclear as written, starting with "within this framework".  Also isovalue hasn't been introduced as a number yet.  So maybe "If we choose an isovalue, c, then within this framework, the set of points v such that f(v)>c is a bounded set---for the example above, a sphere whose radius depends on c.  The default isovalue is c=1.  Increasing the isovalue shrinks..."
5. The text "all of the built-in functions accept these named arguments, which are not repeated in the list below" was reportedly confusing as "list below" was ambiguous.  I'm not loving rewordings I've thought of.  Maybe that paragraph should just go below the list and lead with "In addition to the dimensional arguments listed above, all built-in metaballs accept these named arguments".  That actually feels like it might be a more natural presentation order anyway.
6. "decreases from the metaball center..." I think should maybe say "drops below the isovalue (default 1) somewhere in your bounding box".  I'm not sure that "decreasing from the center" is actually a requirement?  But if it doesn't go below the isovalue in your bounding box, you're going to just get a cube.   (And the current statement would allow that it goes below the isovalue at infinity but not actually in the bounding box.)  Then continue with "If you want your balls to behave similarly to the standard balls it should fall off as 1/d with distance.  See example XX below for how to make a ball that works with cutoff and influence like the built-in balls"
7. Start the first custom metaball example (that doesn't exist yet---the lobe one) with "Custom metaballs are an advanced technique where you define your own metaball shape by passing a function literal <linked> that takes a single argument, a coordinate in space.  It is called dv here, but can be given any name.  Note that in this example we compute a modified distance and then return the inverse of that distance.   The scale factor of 3 hard-codes the size of the ball.  
8. Is it worth after the multilobe example I already wrote, having a repeat of that example but with a helper function and pass-through variables for lobe count and radius, but without cutoff or influence?   That could serve as a friendly stepping stone and the example could feature two interacting multilobes with different parameter radii and lobe counts.

adrianVmariano:
```
function super(x,y,z,s1,s2) =
   (abs(x)^(2/s2)+abs(y)^(2/s2))^(s2/s1) + abs(z)^(2/s1);
isosurface(function (x,y,z) super(x,y,z,.95,.5), 1, voxel_size=.05,closed=true, bounding_box=[[-1.1,-1.1,-1.1],[1.1,1.1,1.1]]);
```

produces extra faces:

image

amatulic:
> produces extra faces

And it probably should. Everything in your picture would be purple if you view with "Thrown together". That would mean the values outside the shape are bigger than the values inside. If you set an isovalue range [0.9,1,1] it works.

adrianVmariano:
I don't understand.  Everything purple would be an understandable solution, but why does that mean I get extra faces?

amatulic:
Finished items 1-7 and 9,11,12

> 8. Is it worth after the multilobe example I already wrote, having a repeat of that example but with a helper function and pass-through variables for lobe count and radius, but without cutoff or influence?

I think that would be confusing. It still confuses me sometimes when I look at the code. I think it's enough just to say that `spec` should specify the custom function as `function (dv) my_func(dv, ....)`

I have already included that example a second time, showing the additional parameters.

> 10. Should the example actually parallel the internal functions more and use a function to return the function literal?

As above, I think that would be confusing.

> 13. Jordan is surprised that isosurface takes 3 args (x,y,z) instead of a vector point. He notes that OpenSCAD syntax for a point is a vector, e.g. translate([x,y,z]) not translate(x,y,z).

Not done yet.

> 14. Add some bounded examples to isosurface, (e.g. superellipsoid?)

TBD

amatulic:
> Everything purple would be an understandable solution, but why does that mean I get extra faces?

I am not sure, and it's curious why it happens only at that voxel size. I note that you have isovalue=1, and your expression evaluates to 1 in places outside the shape but within the bounding box (like at [-1,0,0]. The voxel bounding box got expanded and centered around your original, so it's  [[-1.05, -1.05, -1.05], [1.05, 1.05, 1.05]].

adrianVmariano:
Regarding my proposed suggesting of a second copy of metalobe with a passthrough function, why would that be confusing?  I think it would be LESS confusing than jumping to the noisy sphere example, which uses a passthrough function with two strikes against it:  you didn't just see a simpler version of the same example and it has the additional complication of cutoff and influence in the mix.    I guess I'm not sure what you did because you do say you "already included the example a second time with additional parameters".  

It seems to be that saying "we aren't going to show this because it's confusing" is the wrong thing if by confusing you meant "difficult" or "advanced".  If it's just a convoluted example that doesn't have a point to make, then sure it should be cut.  But if it's a tricky technique that might be useful to users and it's hard to understand then it's **more** important to include examples of how to use that technique.

adrianVmariano:
> > Everything purple would be an understandable solution, but why does that mean I get extra faces?
> 
> I am not sure, and it's curious why it happens only at that voxel size. I note that you have isovalue=1, and your expression evaluates to 1 in places outside the shape but within the bounding box (like at [-1,0,0]. The voxel bounding box got expanded and centered around your original, so it's [[-1.05, -1.05, -1.05], [1.05, 1.05, 1.05]].

How is [-1,0,0] "outside the shape".  The shape should have radius 1 and all six axis points are part of the shape.

amatulic:
I didn't say your lobe example was confusing. I already added it, first the basic one, and then the one with more parameters.

What would be confusing is an example that has the extra support functions allowing the user to avoid putting `function (dv) myfunc(dv, a, b)` in the spec, just so he can put `myfunc(a,b)` in the spec instead, like is done for the mb* functions.

adrianVmariano:
Isn't it a natural question, especially where noisy sphere is **almost** duplicating a built-in metaball, how to fully get the same effect?

amatulic:
Sure it's a natural question, but not one that needs to be answered in the examples. "If you want to do it this way, here are the hoops you have to jump through" doesn't strike me as meaningful or helpful. Built-in is built-in, user-defined is user-defined, there is no problem with them being different.

adrianVmariano:
I would say it's counter-intuitive that requesting "closed" adds those faces.  The figure was already closed and no part of it hits the bounding box.   Basically it's adding the second surface at infinity and clipping it to the cube.   We did say that isovalue=c was the same as isovalue=[c,INF] but I didn't understand this meant that it would draw a shell using the surface at infinity.   So I think we need to add a sentence in the docs about this---maybe where the isovalue is discussed?---and also I think we need to have an example where we show it (or something like it) showing up as a cube and then you have a following example where the isovalue range is changed to [-INF,c] and then it works.   The example might work better if the bounding box goes only to 0 along one axis, so you don't just get a cube.  I think that makes it more clear what's going on.

adrianVmariano:
I think part of the reason for being confused has to do with thinking that I'm computing "an isosurface" instead of an object bounded by an isosurface.   So I wonder if maybe changing the first line docs to

Computes a [VNF structure](vnf.scad) of an object bounded by one or two isosurfaces, intersected with a specified bounding box.

Then here's a stab at rewriting paragraph 2:  

The specified isovalue can be a range or a scalar.  When it is a range, [c1,c2], the returned object is the set of points, p, satisfying c1<=f(p)<=c2.  If f has values larger than c2 and values smaller than c1 then the basic result is a VNF with two bounding surfaces corresponding to the isosurfaces at c1 and c2; this is a shell object with an inside and outside surface.   The isosurfaces don't have to be connected, so you can also get a collection of lots of shell-type objects, in some cases even an unbounded infinite series of objects.   If f(p)<c2 everywhere, then no isosurface exists for c2, so the object has only one bounding surface.  This can result in a bounded object like a sphere or torus, or it can result an an unbounded object such as all the points outside of a sphere out to infinity.  A similar situation arises if f(p)>c1 everywhere.  You are allowed to set c1=-INF or c2=INF, which always produces an object with only a single bounding isosurface.  If you want to obtain a bounded object, think about whether the function values inside your object are smaller or larger than your isosurface value.  If the values inside are smaller, you produce a bounded object using [-INF,c].  If the values inside are larger than [c,INF] will produce a bounded object.   When you give a scalar isovalue, this is equivalent to [c,INF], so it will produce a bounded object when the points inside the object are **larger** than the points at the isovalue.  

The description above suggests that the computed shape is a single object, but depending on your function, the object may have many (even infinitely many) disconnected components.  It is possible to have an unbounded object combined with several shell objects.   When your object is unbounded it may display as simply the specified bounding box if `closed=true`.  

That hopefully transitions to the paragraph on the bounding box.  

Then at the end of that paragraph add:

If your object is unbounded then when it is intersected with the bounding box, the result may appear like a solid cube, because the bounding surface is hidden inside.  Setting `closed=false` will remove the bounding box faces and expose the inside structure.  If you want the bounded object, you can correct this problem by changing the isovalue range:  if you had specified `isovalue=c` then use `isovalue=[-INF,c]`.  If you had used `[c1,c2]` then one of `[-INF,c1]` or `[c2,INF]` will do the job.  (Or you may have to inrvoke isosurface twice to get both regions.)

Then we should have examples of multi-component isosurface, and the previously noted unbounded case.  

I actually was wondering if the isosurface behavior is backwards and if the default for a scalar should be [-INF,c] instead of how you did it.  I feel like that's more typically how somebody would construct a (bounded) set.   (Yes, I know this is backwards relative to metaballs.)

amatulic:
That's a lot of words, but they look good to me. My only concern is that the previous words made a distinction between interior and exterior, which is important for a VNF because each facet has an interior and exterior side. I thought it was important to say that by default, "exterior" is wherever f < c.

amatulic:
I removed the bit about vnf_bend.

I propose:
* `closed=true` by default for metaballs, which is a specific case of isosurface where this is appropriate
* `closed=undef` by default for isosurfaces, which are more general-purpose
  * if undef, then assume `closed=true` for any finite range of isovalue, or for the range `[-INF,c]`.
  * if undef, then assume `closed=false` for a scalar isovalue or the range `[c,INF]`
  * if set by the user, then use that.
 * modify the examples to include this argument, since it matters a lot depending on your function

adrianVmariano:
Note that I tossed out the idea about changing the `closed` default for discussion, not implementation.   First of all, metaballs should be closed=true by default.  This was never about changing metaballs.  

The idea was that closed=true can be confusing, so therefore maybe closed=false should be the default.  The problem with the above proposal is that closed=true is potentially confusing when the shape is unbounded, and that can happen with **any** isovalue range.  It's not a special property of [-INF,c].   So having a complicated default just creates more potential confusion while only solving the problem in one situation.   Actually did you change the default range?  Because my example where I encountered the problem was [c,INF].  

Also note that the default for a parameter is not "what it says on the function line" it's how that parameter behaves when you run the function.  You can implement a default by writing `foo=default` on the definition.  You can also implement a default by writing foo=default(foo,value) in a let statement, or by other more complicated means.  So I would more succinctly describe your proposal as "Default: closed is false when c1 is -INF and true otherwise". 

Here we get a cube in each case with each possible isovalue range type:

```
isosurface(function(p) exp(-1/norm(p)), isovalue=[.5,1.5], bounding_box=[[-3,-3,-3], [3,3,3]], voxel_size=.5, closed=true);
isosurface(function(p) exp(1/norm(p)), isovalue=[-INF,3], bounding_box=[[-3,-3,-3], [3,3,3]], voxel_size=.5, closed=true);
isosurface(function(p) exp(norm(p)), isovalue=[3,INF], bounding_box=[[-3,-3,-3], [3,3,3]], voxel_size=.5, closed=true);
```
Might be worth banning the case if `isovalue=[-INF,INF]`.   Existing code could be simplified by doing `isovalue=is_list(isovalue)?isovalue:[isovalue,INF]` and then referring to the [0] and [1] entries.  

After pondering the idea of defaulting closed to open I think I don't really like it.  The justification is that it makes things less confusing.  The cost is that we have a default which is basically wrong most of the time.   I think that places the priority incorrectly.  So I lean towards keeping the default closed=true and hitting the topic in the docs.   Maybe we have a bold face sentence that reads "Why did I just get a cube?" followed by instructions that either the bounding box is too small or the range is flipped.   That, combined with an example that shoes the cube problem and solution seems like it should be good.

Also, I think that it's better to encourage people to know what they are doing---that is, people should think for a moment about the interval that they need.  Making closed=true by default instead supports a trial-and-error approach to getting things right.   This makes me wonder if the scalar default should be removed and only a range permitted.  That way you're forced to think explicitly when picking the isovalue whether you want to go up to INF or down to -INF.  That could decrease confusion around the behavior of scalar isovalue.

adrianVmariano:
1. Consider banning [-INF,INF] for isovalue
2. Consider removing the scalar default for isovalue
3. isovalue assertion can get an error in the assertion test because you never checked that input was numeric.  It's a bit tricky because INF is allowed.  assert(is_finite(isovalue) || (is_list(isovalue) && len(isovalue)==2 && is_num(isovalue[0]) && is_num(isovalue[1])) may be adequate.  (Doesn't rule out NaN as a possibility.)  
4. Would a shorthand bounding box like a simple scalar for a centered box of that size (e.g. the way cube(5) makes a size 5 cube) be worth implementing?  
5.

adrianVmariano:
It appears that some closing faces are missing here:

```
isosurface(function(p) (p.x^2-p.y^2)*p.z+p.z^2, isovalue=[-5,15], bounding_box=[[-5,-5,-5],[5,5,5]], voxel_size=.5);
```
image

adrianVmariano:
```
isosurface(function(p) (p.x^2-p.y^2)*p.z+p.z^2+p.x*(p.y^2-p.z^2), isovalue=[-INF,5], bounding_box=[[-5,-5,-5],[5,5,5]], voxel_size=.5);
```

image

adrianVmariano:
A bounded example?

```
isosurface(function(p)  (p.x*p.y*p.z^3+19*p.x^2*p.z^2)/norm(p)^2+(abs(p.y)^3+abs(p.z)^3+p.x^2), isovalue=[-INF,35], bounding_box=[[-8,-5,-5],[8,5,5]], voxel_size=.25);
```

Wonder if it can be simplified.

image


or perhaps

```
isosurface(function(p)  (p.x*p.y*p.z^3+19*p.x^2*p.z^2)/norm(p)^2+norm(p)^2, isovalue=[-INF,35], bounding_box=[[-8,-8,-6],[8,8,6]], voxel_size=.25);
```

image

adrianVmariano:
Could this appear in a model as a kind of support pillar? 

```
isosurface(function(p)  (p.x*p.y*p.z^3-3*p.x^2*p.z^2)/norm(p)^2+norm(p)^2, isovalue=[-INF,35], bounding_box=2*[[-10,-10,-8],[10,10,8]], voxel_size=.5);
```
image

Would be better if it could be adjusted so the cut face is entirely on top.

amatulic:
Finished all items 1,3,4,5. The last one I had already done yesterday.

> Consider removing the scalar default for isovalue

There is no scalar default for isovalue in isosurface(), in fact there has never been a default at all. We already agreed that metaballs should have a default isovalue=1.

I like those examples. The second one looks most interesting.

Working on debugging those holes you found in the clipping surface.

adrianVmariano:
By scalar default I meant that you're currently allowed to give a scalar for isovalue and it defaults to [c,INF].   The proposal is that ONLY a range is accepted (scalar is an error) and you enter [c,INF] or [-INF,c] if you want an unbounded interval.  This was mentioned with more context earlier in the longer message about default for closed.  

Yeah, I tweaked the last example and actually posted the example on the chat with it attached to some cylinders and a cube.   For some reason when I tried to enlarge the bounding box I didn't go about it right and it didn't seem to be helping.  But if you just lower the box until it's below the flat face that obviously will work.  

I do wonder if there's a way to design forms with more intent than my approach, which is to sort of randomly stick equations together.

amatulic:
I fixed the holes by changing one "<" to "<=". The holes appeared when one corner of a clip face is exactly equal to max isovalue.

I am not sure of the purpose of requiring isosurface() to take a range for isovalue. Perhaps just show a range in all the examples, describe isovalue in terms of ranges in the documentation, and mention that a scalar is the same as [c,INF]? And what about metaballs? It can take either a scalar or range. A scalar makes sense probably for 99.9% of what anybody would want to do with metaballs, but now and then I'd like to make a metaball shell.

adrianVmariano:
Remember my caution about the use of `=` with floating point?  The question is whether comparisons should be <x instead of <=x or if it should be <x+eps.  Do you understand why it should be strict < instead of <=?  

The reason I proposed having isosurface require a range is that a scalar is ambiguous in that it could be [-INF,c] or it could be [c,INF].  Now sure, you had it defined to be the latter---so strictly speaking, it's not ambiguous.  But by accepting a scalar you enable the user to not think about this distinction---like I did---which makes it more likely to lead to confusion where you get the shape backwards because you never had to make a choice about which way it goes.  It seems likely that if I had needed to specify a range for that original test that I got backwards that I would have gotten it right.  

Metaball should definitely still accept a scalar, and range also is OK.

amatulic:
It needs to be <=. These are comparison of function values at the corners. If the corner is equal to the max isovalue, the regular surface triangulation was picking this up but the clipping surface wasn't. If it isn't equal, even off by eps, it would still work. Also, f+eps does not work when f has a high value.

amatulic:
All tasks in this PR are done, examples have been added, docs changed as discussed. This should be the "final draft" if all OK and docsgen doesn't abort.

amatulic:
Please review and merge if it looks OK.
* Added new args `grow_bounds` and `show_box`
* Added triangle area check to final step of isosurface()
* Made all voxel corner function comparisons <= isovalue rather than < (it was < to avoid zero-area triangles, which is no longer a problem)
* Added example showing effect of low influence. Actually the fingers of the hand model show this, too.
* Removed low influence from most elephant parts and substituted cutoff to smooth it out a bit.

adrianVmariano:
Probably not going to have time to actually review until tomorrow.  One thought, it seems like grow_bounds arg should be instead exact_voxel or fixed_voxel or something like that.  Nobody is going to think: I want the bounds to not be what I asked for.  The reason you want the bounds to grow is because that happens when you don't allow the voxel to change and you want to prevent he voxel from changing.

adrianVmariano:
Should include an example with artifacts for isosurface and then show taking log to smooth it out?

amatulic:
I made the changes you suggested.

I tried to demonstrate artifacts with a tilted cube but they aren't strong artifacts. Here's an example:
````
include <BOSL2/std.scad>
include <BOSL2/isosurface.scad>

function tiltcube(point, a, size, n) =
    let(p = rot(a, p=2*point/size))
        1 / (abs(p.x)^n + abs(p.y)^n + abs(p.z)^n) ^ (1/n);

isosurface(function (point) tiltcube(point, [0,15,22], 7, n=50),
    isovalue=[1,INF], voxel_size="auto",
    bounding_box=10, show_stats=true);
````
The artifacts are faint:
image

What did you do to get those pronounced ridges?

**Comment by adrianVmariano** [comment]
The cube artifact example was something like:

```
metaballs([zrot(22)*yrot(15), mb_cuboid(5, squareness=0.9, influence=0.05)], voxel_size=.5, bounding_box=12);
```

The exact artifacts are pretty sensitive to the voxel size.

amatulic:
I saw what I was doing wrong with the cube, I wasn't taking it to the power of "influence". Anyway, the example in chat is better. I didn't include it in the last push yet, will do so later tonight. Basically the rule is, if you take the log of your function, you must also use the log of your isovalue. And, I learned, it must be the same log, not ln(function) and then log(isovalue) -- I did that by mistake.

adrianVmariano:
I generally think it works better to use two separate examples, which makes it easier to see which code produced which result.  And when the pictures are small, it can work better, especially if you can get the code to be on the right in the wiki.   In this case, the difference is huge so it's not like you need to study them carefully side by side.  But I'll let you be the judge on what works best.  

We should mention, if it wasn't already that the isosurfaces are preserved by taking the log (as long as the function is positive everywhere), so taking the log in principle gives exactly the same result.  f(p)=c ===> log(f(p)) = log(c)

I will look over the revisions in the morning.   The one question I have about artifacts is whether they can look different under different circumstances.  Like the artifacts on the mb_cuboid look different from the artifacts on the above thing.   So we may want to mention that the artifacts can manifest in different ways and any case of unexpected lack of smoothness that is not improved by smaller voxels might be caused by nonlinearity of your function around the isovalue.   (Maybe you did this already.)

amatulic:
It's obvious in this example which line produced which result. There's a left and a right in the image as well as in the code. Putting them both in the same place makes it clear that the underlying function is identical for both cases.

The artifacts all have one thing in common, as far as I've observed: They are rounded ridges. The spacing and amplitude of the ridges vary depending on voxel size, isovalue, and the underlying function.

I reformatted the example code so that it should fit on the right side of the image. I'm going to wait until tomorrow to see if @RAMilewski provides any further updates in #1559 and include those with the final submission after your review.

adrianVmariano:
Comments on isosurface() description:

1. Add to this "Setting isovalue to `[-INF,c_max]` or `[c_min,INF]` always produces an object with a  single bounding isosurface."  sentence ", which could be unbounded".  
2. Not sure if you made the change I suggested of exact_voxel=true instead of fixed_bounds=true, or if you didn't like that proposal.  It seems like varying voxel size should be the default with fixed bounds because that will be the prevailing user expectation.  
3. Looking at the voxel_size/auto_voxel API I think it needs to be redone as follows:  two arguments should be voxel_size and voxel_count.  Only one can be given and there is no default.  That eliminates the need to give voxel_size="auto" and then also giving the count argument in auto_voxel.  Assert an error if the user provides both parameters.
4. You never say that voxel_size can be either scalar or vector.   Probably add that right after it's introduced in paragraph 3.  
5. Should there be an example of a very non-cubic voxel size, like voxel_size=[1,1,10]?   Is there any reason to do this or not do it?  (Maybe you have a very long skinny object?)
6. When describing the behavior of fixed_bounds or a replaced name for it I would not put one of the alternatives in parentheses but rather give them equal weight.   
7. "causes the VNF to end at the bounding box" may suggest that it gets clipped if one thinks of a VNF as a solid.  Maybe "causes the VNF faces to end at the bounding box?"    
8. It seems that you don't document that the bounding box can be a scalar in the text.  
9. voxel_count (or "auto") feature appears to be undocumented.  Oh, wait...it's documented in a separate place from voxe_size.  That's confusing.   I think a lot of users won't read the paragraph that talks about voxel size and run time because they'll dismiss it as obvious.  
10. First example seems kind of wordy and confusing because it seems like r is both the name of the function and the isovalue.  I'm not sure what "the isosurface at r exists at every point where the expression equals r" means.  Is "isosurface at r" meant to refer to isosurface with isovalue r?   Also objects exist or they don't exist.  Objects don't exist at a point.  So whatever's happening at every x,y,z point it should be something other than "exist".  
11. Second example says "if we the isovalue".  Add "set"?  But also could be less wordy.  It says "Example 2" already.  You don't need to repeat that in the text.  So "An isovalue range of [8,10] gives a shell with inner radius 8 and outer radius 10".  (I think interior and exterior are technically the wrong words here.)  
12. I would start example there with "Here we set the isovalue range to [10,INF]...."  Then at the end it doesn't make sense to cut an isovalue from something, since that's a number.  It's hard to see a decent way to rephrase that while being technically correct.  I got the awkward:  "with the bounded object bounded by the isovalue 10 cut out of it" when I tried to do that.  I suggest instead something like "appears as the bounding box with an interior hole whose boundary is the sphere with radius 10".  
13.  When introducing the gyroid maybe add "unlike the sphere, the gyroid does not have a bounded isosurface"?  Or something about how it's always unbounded.  You might consider if you think the order of examples is the best natural order to present them in.  I was wondering if gyroid should be later, but I'm not sure.  
14. I don't see any examples of artifacts yet?  and use of log?  I guess that's still coming?  
15. Does auto voxel computation round up or down?  This is not documented.  If I ask for 8000 do I get no more than 8000 or am I guaranteed 8000?
16. I suggest generally not being parsimonious with examples.  Consider including all the examples I listed above unless you really think some of them look the same, because this gives more of a flavor of what's possible.   (The first one makes me think of a finger shaped recess if you just use part of it, though there's the problem of how to round the top.) 
17. The show_box option fails with errors if the box is given as a scalar

For metaballs docs many of the above comments also apply, so check that.  Also
1. I asked previously about talking about metaballs being summed.  You added this to paragraph 3, but this is a technical detail that makes no sense to the reader at this point.  It belongs in the section on custom metaball functions.  
2. paragraph 3 introduces bounding box but doesn't mention the option of scalar size.   Instead of "outside the bounds" I think "outside the bounding box".   Also voxel_size should be noted as a scalar or vector when introduced.  
3. Does the problem with faces that lie on the bounding box even still occur?  I could not reproduce it.  Maybe fixed by the equality test adjustments?
4. "availableto" -> "available to"
5. In paragraph on influence should we add "very small influence values can produce ridge-like artifacts or texture on the model.  To avoid these keep influence above about 0.5 and consider using cutoff instead of very small influence"
6. How easy is it to swallow a metaball with a negative ball?  Can this be done with two balls that don't share a center?  A brief test suggests the answer is no.  `metaballs([IDENT, mb_sphere(r=4), right(.1), mb_sphere(r=6,negative=true)], bounding_box=2, voxel_size=.05);` produces a small very artifacty ball not improved by taking logs.  
7. For the isovalue paragraph, maybe lead with "The isovalue parameter parameter applies globally to all your metaballs and changes the appearance of your entire metaball object, possibly dramatically.  It defaults to 1 and you don't usually need to change it.  If you increase the isovalue..."   Then at the end of this paragraph add something like "More details on the precise meaning of the isovalue appears below in the section on custom metaball functions"
8. "Both major and minor radius/diameter must be specified regardless of how they are named." is confusing because it suggests that you have to use r_maj or d_maj and also r_min or d_min.  I think a statement like "You must provide a combination of inputs that completely specifies the torus" would be better.  
9. "Cutoff is measured from the object's center unless otherwise noted below"  The "below" needs to change to "above".   Is it worth noting that cutoff for capsule is measured from the line of the center, not the origin center?   Does anybody know how torus cutoff works?    
10.  "defines a bounded set &mdash; for example, a sphere", the dash should be a colon
11.  There seems to be as yet no discussion of artifacts or artifact examples.

amatulic:
Thanks for going through this in detail.

We have disagreement about a logical list of function arguments versus a good user experience in actual practice. Those two goals don't always align.

Isosurface()

1. Done.
2. What used to be `grow_bounds=true` is now `fixed_bounds=false`. Just a change of argument name with opposite logic. The user expectation is for bounds to be an arbitrary workspace. That's how it's intended to be used, and that's how it **is** being used in all our back-and-forth experiments. For example, I notice your own bounds are almost always bigger than needed, and approximate. You don't really care what the bounds are as long as they contain the feature of interest. The user typically adjusts bounds to optimize speed. I held my nose and added this capability to have unchanging the bounds although I strongly disagree with it, because the suggestion originally arose without the benefit of experience actually making metaballs or isosurfaces. The universal expectation, in all I have read, and in my own tests, is for voxels to be cubes. Bounding box is just a work volume. Anyone who really wants to hold the bounds fixed can either set `fixed_bounds=true` or set it up so that the voxels fit inside.
3. Not done because it isn't a good user experience. It's `voxel_size="auto"` because I already tried it the way you suggest and found it inconvenient and impractical. If I am experimenting, I don't want to be changing an argument name back and forth. Sometimes I have `voxel_size` as a variable in my code, and I just want to change its value. In practice, I want to see a rough picture with an automatic voxel size, and then change the voxel size without having to switch to a different parameter name. I am really enjoying the `voxel_size="auto"` feature now. I am using it a lot. It wouldn't be as convenient if I had to use some other argument just to see a "draft" version.
The size of the voxel is paramount, it controls resolution. The total number of voxels in an arbitrary bounding volume is irrelevant to me. A parameter like `voxel_count` would confuse me because I would think it's the count of voxels in my shape. The total voxel count isn't something I would **ever** care about. I called it `auto_voxels` because this is a good default for quick rough views when voxel_size="auto" and I doubt I would ever need to change the default.
4. The options for voxel_size are already documented in the argument description, but I put it in the text too.
5. Done, added example showing where it would be necessary to change the shape of the voxel.
6. Done, removed parenthetical note, actually gave fixed-size bounds more weight by starting out "Alternatively,...."
7. Done.
8. The scalar option for bounding box is already documented in the argument description, but I put it in the text also.
9.  It isn't clear what's missing here. `voxel_size="auto"` and `auto_voxels` are always mentioned together, wherever they occur, and are already documented in the paragraph about execution time, as well as in the argument descriptions about `voxel_size` and `auto_voxels`. As I said in (2) above, obody needs to care about the number of voxels. I doubt anyone would bother setting the `auto_voxels` argument. It just needs to be a value that results in a reasonable execution time.
10. Done. I agree it was confusing. I reworded example 1.
11. Done.
12. Done, reworded and simplified as suggested.
13. Done.
14. Already done. The artifacts and use of a log is the second-to-last example.
15. How this works isn't so simple. The number of voxels you get is *approximately* what `auto_voxels` requests. This is documented in the argument description for `auto_voxels`. As I said in (2) above, the exact number shouldn't matter to anyone. Here's how it works: A cubical auto-voxel size is first calculated as `cube_root(bounding box volume / number of voxels)`, but we don't know if these cubes fit exactly in all three box dimensions. Likewise, if the bounding box volume is fixed and the auto-voxel cube is resized instead, we don't know if the total number of resized voxels is *exactly* the amount of voxels requested. It's *possible* for it to be exact because the default is 8000 voxels or 20^3.
16. Already done, I think. Not sure which example you're referring to. The finger-shaped recess is the example demonstrating the use of a scalar for a bounding box. And the support pillar example is used to demonstrate `show_box`.
17. Fixed, I noticed that last night after pushing the PR. But I wonder why the docsgen check passed with that error?

Metaballs() - applied relevant changes above here too.

1. Done. Originally I wasn't sure where that sentence should go.
2. Done.
3. Yes, the problem with flat surfaces coplanar with the bounding box still occurs. Worst case is you have some degenerate faces that got generated because they are part of the isosurface as well as on the side of the bounding box.
4. Done, and I found a couple other typos also.
5. Done, great suggestion, I just added an example but didn't think of describing it in the text.
6. I removed the part about swallowing other metaballs, changed it to "...which can result in hollows, dents, or reductions in size of other metaballs." The only time I've seen a metaball swallowed is when its infinite center is inside a voxel, in which case all 8 voxel corners around it are finite, and can be reduced to less than the isovalue.
7. Done, that's better wording.
8. Done.
9. Done, fixed wording. Does cutoff on a capsule affect an object near the end of the capsule, along the capsule's axis? I believe a cutoff with a torus is simply a torus-shaped cutoff, the shape being similar to what you'd get if you reduced isovalue. You'd get a fatter torus, possibly with small dents in place of a hole if it gets too fat.
10. Done.
11. Example 14 demonstrates artifacts with metaballs. I have mentioned this example in the text you suggested about artifacts.

PR #1577, opened 2025-02-26
Anachronist isosurface changes

amatulic:
* Changed isosurface `f` to accept x,y,z args instead of a vector
* Moved paragraph about duplicate points out of metaballs() and isosurface(), moved to the top of the page, added comment about Manifold generating warnings to that paragraph.

Re-tested all isosurface examples after modifying them for x,y,z args.

### DIFF
--- a/isosurface.scad
+++ b/isosurface.scad
@@ -35,7 +35,16 @@
 //   isosurface is unbounded and periodic in all three dimensions.
 //   .
 //   This file provides modules and functions to create a [VNF](vnf.scad) using metaballs, or from general isosurfaces.
-//   
+//   .
+//   The point list in the generated VNF structure contains many duplicated points. This is normally not a
+//   problem for rendering the shape, but machine roundoff differences may result in Manifold issuing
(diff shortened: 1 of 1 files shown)


adrianVmariano:
It may be worth writing the paragraph about multiple points two or three times, once at the top and in the docs repeated in both metaballs and isosurface.  I think a lot of users won't see the top text if they jump right to the manual for the function they are using, so I would say that at the top is the least important.  I think of the  top text as summarizing what this file does so you can decide if it is useful to you or not rather than trying to give details about how to use it.  

Another possible idea might be to write an entire doc section about how isosurfaces and metaballs work and what some common characteristics are, sort of like gears.scad has an intro to gears section.  (Not proposing that this happen now.)  

You might want to write 1e-8 or whatever instead of 0.00000001 because the latter is hard to read.

amatulic:
I copied that paragraph back into where they were. quantizing to 1e-8 didn't fix the problem in Richard's example, but 1e-7 worked, so I recommend 1e-6, which is the level of precision isosurface uses to judge two values as equal.

PR #1580, opened 2025-02-28
fixed mb_octahedron size, fixed doc issues

amatulic:
Captured changes requested from chat.
* Fixed mb_octahedron size depending on squareness by changing `r` by a cubic correction factor.
* Tweaked zooming and positioning of a number of examples
* Fixed major and minor doc issues (major was: usage args out of order, wrong args in mb_capsule and mb_connector).

### DIFF
--- a/isosurface.scad
+++ b/isosurface.scad
@@ -1275,17 +1275,17 @@ function _mb_octahedron_influence(point, invr, xp, ex, neg) =
    let( p = point*invr,
         dist = xp>1100 ? abs(p.x)+abs(p.y)+abs(p.z)
         : (abs(p.x+p.y+p.z)^xp + abs(-p.x-p.y+p.z)^xp + abs(-p.x+p.y-p.z)^xp + abs(p.x-p.y-p.z)^xp) ^ (1/xp)
-    ) neg * (r/dist)^ex;
+    ) neg/dist^ex;
 function _mb_octahedron_cutoff(point, invr, xp, cutoff, neg) =
    let( p = point*invr,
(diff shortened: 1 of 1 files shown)


adrianVmariano:
Just from looking at the code, I think there's a lingering "r" value that will be undefined in mb_octahedron for every case but basic.  

I'm not sure why you didn't just modify r and leave the code like the non-basic case currently looks---that is in principle faster since it's a scalar product at that point instead of vector.   

That ugly cubic approximation should be replaced by the exact calculation, which looks to be `(3^xp+3)^(1/xp)/3`.   I just plugged the point [1/3,1/3,1/3] into the octahedron equation.

amatulic:
OK, I fixed octahedron simply to get the radius from _mb_octahedron_basic and correct it.

Good catch about the extra r term in the non-basic functions, those shouldn't have been there. I think I may have copied mb_cuboid to make the new mb_octahedron and didn't correct the final line. Corrected, and tested.

Currently octahedron uses r as the default size. I notice that the BOSL2 octahedron uses tip-to-tip size, not center-to-tip. I'll change it and push another one, along with the other corrections.

adrianVmariano:
Yeah, I suppose just calling _mb_octahedron_basic also works.  

It looks like BOSL2 octahedron uses a "size" parameter and accepts a vector.  Not sure if it makes sense to support that, e.g. the same way it's done in mb_cuboid.

amatulic:
I just fixed mb_octahedron to support a vector size, and it's working.

PR #1587, opened 2025-03-04
Fixed clip face edge cases, added debug view to metaballs

amatulic:
* Replaced original algorithmic handling of bounding box clipping triangulation with new code that properly accounts for more than 2 isosurfaces intersecting a face. Tested extreme cases, compared with public examples.
* Added debug view to metaballs() function
  * **No incompatibility with existing API** so there's no hurry to update the wiki.
  * New argument `debug` has been added to metaballs() module.
  * New argument `nodebug` has been added to each mb_ function to suppress debug display of a specific metaball that may be covering up other more important details.
  * Duck metaball example added.
  * Documentation updated for debug view.
  * All examples tested

**Notes about clipping triangulation**
(Added here for posterity because I can't put this image in the code comments.)

For marching cubes, there are 256 voxel configurations depending on whether a corner is above or below the isovalue. Each of the 8 corners has two states, so the index of a cube in a triangulation lookup table is an 8-bit binary number. An isovalue range is handled as two separate isosurfaces triangulated independently.

Where the bounding box and a voxel share a face, however, both extremes of the isovalue range must be accounted for at once, on that voxel face. A voxel face has 81 possible states in which corners can be below, inside, or above the isovalue range. Because each corner has 3 states, the index ID of a face is a 4-bit base-3 number ranging from 0000 to 2222 (81 values).

Of those 81 states, there are 24 states excuding rotations. Four states are mirrored but I am not considering mirrors as equivalent. That gives 24 possible triangulations, which are rotated as needed for all 81 states. The figure below shows these 24 states, labeled with the decimal index, base-3 index, and number of versions with rotations.

image

Ambiguities arise when two diagonally-opposite corners have the same state and the two other corners have a different state. There are different ways to triangulate. The convention used with the marching cubes triangulation is to assume that the value in the middle of the face is somewhere between the two lower diagonal values. In elevation terms, there is never a ridge connecting two diagonally high corners; rather, a valley connects the low corners. The same convention is used here for compatibility.

### DIFF
--- a/isosurface.scad
+++ b/isosurface.scad
@@ -625,7 +625,6 @@ _MCTriangleTable_reverse = [
  []
 ];
 
-
 /// _cubindex() - private function, called by _isosurface_cubes()
 /// Return the index ID of a voxel depending on the field strength at each corner exceeding isoval.
 function _cubeindex(f, isoval) =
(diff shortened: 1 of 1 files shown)


adrianVmariano:
I wouldn't bother with "This duck is used as a basis for the examples to follow."  

2nd one could be tightened to:  "Specifying debug=true shows the elements used to construct the duck.  Positive balls...etc....unfortunately.....   Note also that removing the voxel_size parameter from metaballs speeds up the preview."'

You don't want to lead with the text about voxel size, since that's the least important thing about the example.  

Then final example could start:

"Adding no_debug=true to the skull metaball function suppresses its display and reveals..."

Did you consider making the debug balls transparent?  

Is the passing of VNFs for the debug shape documented somewhere?   I didn't see it.

adrianVmariano:
I think it's better to document things than to fear that your docs are too long.   I was forced to greatly lengthen the path_sweep docs for example because of complaints about confusion and inadequacy.   The one alternative I see is to break out as much as possible into a separate section about the general workings if metaballs and isosurfaces so that the docs can be shorter (with less repetition between the two cases).  

The ability to debug should definitely be mentioned in the main description section.  

I don't think  you necessarily need to include an example of passing a debug VNF but it might be good to document how it works in the section on custom functions.  You can add another paragraph.  

Here's a wacky example of nested colors and an apparently colored % object:

```
recolor([1,0,0,.5])ghost_this() cuboid(10)
  recolor([0,1,0,.5]) cuboid(5);
```

I'm not even sure I understand what's going on there.  The shade of the middle object changes oddly with view.   But in any case, the point is made that you can't see a transparent object that is inside another one, so it wouldn't help with hidden debug balls.

amatulic:
I wonder why the `#` marker seems more reliable for transparency? It's an ugly red color, but it shows objects inside and through the transparency. That's what I use mostly, when I want to make something transparent.

revarbat:
Because # marked items are drawn _after_ everything else, then everything else is already drawn, and the # marked items can be drawn over them with partial opacity.

adrianVmariano:
Are you comparing # to % and saying # is more reliable?  The % "background" marker is weird because it takes the object out of the model.  So it doesn't participate in a difference or intersect operation, for example.  The # highlight marker does not do that.  The text on the wiki says it creates the model and then displays the object in red afterwards.  Both seem to be transparent to everything regardless of ordering.

amatulic:
The main changes to the documentation are:
* New paragraph starting "The spec parameter is flexible" with a bullet list showing the four possible forms of the function specification
* New section with heading "User-defined functions in debug view"
* Addition of the last example with its explanation

See the [diff for this specific commit](https://github.com/BelfrySCAD/BOSL2/pull/1587/commits/c7ad45021b2847135885d29b5365194f2169a206).

adrianVmariano:
"lternately, `voxel_count` may be specified to set the voxel size according to the requested
//   count of voxels in the bounding box."

you can set `voxel_count` to fit approximately the specified number of boxels into the bounding box.

" so if objects seem to be missing, try making `voxel_size` smaller."

try making `voxel_size` smaller or `voxel_count` larger.  

It seems odd to have a special heading for user defined functions in debug mode.  I don't think that heading is necessary.  (Generally speaking I wouldn't think a heading is ever necessary for a single paragraph.)  

"A function literal in the form `function (point) custom_func(point, arg1, arg2...)` wh"

This text should say "a function literal accepting a single argument" which I think should be described as "a 3-vector giving a point in space relative to the metaball's center."  The current description as a "distance" is wrong---distances are scalars.  The text that says "point is supplied internally" is also weird.  What matters is what the user needs to supply, a function that accepts one argument, a point.    

The description you have given above muddies the water.  That's not what a function literal is.  That's one particular way of constructing a function literal.  The user can supply the function literal any way they want.  It could be stored in a variable f and the user could simply write "f" at that point.  They don't need to write all that stuff.   And it's very unclear what arg1 and arg2 are supposed to be.  (Obviously I know what they are, but a user who doesn't understand this stuff is going to be saying WTF are arg1 and arg2 and where are they set.)  

If you *really* think it's necessary to stick basic openscad tutoring into the metaball docs (which you were complaining were too long...) then you should probably create a separate paragraph about how function literals work.

amatulic:
Made changes as you suggest. You're right, we don't need a tutorial on function literals.

If a paragraph has a completely different topic than the previous heading, it gets its own heading. I've renamed the heading and expanded it into two paragraphs about debug view, with one topic being the default VNF for custom functions.

adrianVmariano:
The idea that you need a heading because the topic changed seems bizarre.  You have a new **paragraph** because the topic changed.  But whatever.  

Is there interest in a tetrahedron metaball, if we can figure one out?  I mean, we know how to make that, just maybe not how to do it efficiently.    Might be more efficient to make a non-regular tetrahedron.   If this is a hope for the future you might not want to make debug spheres tetrahedral.

amatulic:
If you use headings, then to write about a different subject after one heading, you need a different heading. That's the case here, and anyway it's now two paragraphs about debugging.

A custom metaball is unlikely to be a sphere. If the default custom shape is a sphere, the only  way to distinguish it from all the other spheres is with color, which may be obscured somewhat by the transparency overlay. So I wanted it to be a more unique shape that's unlikely to be used if we ever implement it as a metaball. The tetrahedron is a more efficient shape in terms of generation and memory too.

It looks like this:
image

I don't understand why the rounded octahedron starts out as tetrahedral arrangement of points. If I use an almost identical function that gets larger outward instead of inward, the isosurface is a tetrahedron, but inverting the output makes it an octahedron? That was serindipitous and allowed me to replace the sharp-edged octahedron to be more consistent with all the other rounded shapes, but I still don't get *why* it's an octahedron.

I can imagine using a stretched tetrahedron in place of a cone for the giraffe head, if such a metaball exists.

In the unlikely case in which a user declares the spec as `[trans, [function, [sign,vnf]]`, but then specifies `[]` or `[sign, []]` after the function, that's now fixed. In this way you could specify a sign for the default VNF if you don't feel like making a default, and it becomes a color instead of gray.

adrianVmariano:
My only remark on using the tetrahedron was to pose the question: what is the chance we implement mb_tetrahedron?  If it's significant, maybe you don't want to use it for some other thing?  But you can decide that.   There is a conceptually simple way to implement tetrahedron, identical to how mb_cyl works but in 3d.  But it seems likely to be slow

adrianVmariano:
"An array containing a function literal and a debug VNF, as `[custom_func(point, arg1,...), [sign, vnf]]`, "

I think that should just say `[custom_func, [sign,vnf]]` and is confusing as written.  

"A function literal accepting as its first argument a 3-vector representing a point in space relative to the metaball's center."

I would write just "accepting a 3-vector..." or maybe "A single argument function literal accepting..."

I mean, yeah, you can pass a 12 argument function, but since the extra 11 arguments cannot be used for anything and will just assume their default values, there's little point in it.  So offering that option to the user is just a potential source of confusion.  

I wouldn't have bothered with this minor tweak but since I suggested the above:

"`$metaball_vnf` is available to child modules to get the VNF of the metaball scene."

All side effects are available to child modules...that's what a side effect is...so that's redundant.  And "to get" is sort of vague.  How exactly does one get the VNF then?  Like is it a function literal that you call to obtain the vnf?  I would simply say "$metaball_vnf is set to the VNF of the metaball scene"

amatulic:
> what is the chance we implement mb_tetrahedron?

Not very good, I'd say. I managed to make one because I was annoyed that I couldn't, but it isn't well behaved, even when I do the scaling trick you suggested with the octahedron to make it fit an inscribed sphere of the given radius.

adrianVmariano:
It seems like going to barycentric coordinates might be helpful for the tetrahedron.  Probably would simplify the calculations for determining distance to a tetrahedron, at the very least.  

It this ready for merge?

amatulic:
Ready to merge. No hurry on updating the wiki. None of the changes are incompatible with the existing online documentation.

amatulic:
Likely not spheroform, but was simple to make.
image
Unstable with exponent>300. Goes to a sphere at lower exponents. Not very useful though.

PR #1595, opened 2025-03-11
New features: 2D contours and 2D metaballs

amatulic:
Isosurface.scad updates:
* Added contour() along with supporting functions and initializations.
* Added metaballs2d() along with supporting functions
* Added 2D metaball functions for circle, rect (squircle), trapezoid, stadium, connector2d, ring
* Included 2D metaball examples by adapting some of the existing 3D examples
* Included a couple of contour examples

Opening this PR as a draft for now.

### DIFF
--- a/isosurface.scad
+++ b/isosurface.scad
@@ -1093,7 +1093,8 @@ function _contour_pixels(pixsize, bbox, fieldarray, fieldfunc, pixcenters, isova
                     // get center value of pixel
                     if (pixcenters)
                         is_def(fieldfunc)
-                            ? fieldfunc(x+hp, y+hp) : 0.25*(f0 + f1 + f2 + f3)
+                            ? min(1e9,max(-1e9,fieldfunc(x+hp.x, y+hp.y)))
+                            : 0.25*(f0 + f1 + f2 + f3)
                 ],
                 pixcoord = [x,y],
                 pixfound_isoval = (min(pf) < isovalue && isovalue < max(pf)),
--- a/isosurface.scad
+++ b/isosurface.scad
@@ -35,14 +35,14 @@
 //   isosurface is unbounded and periodic in all three dimensions.
 //   .
 //   This file provides modules and functions to create a [VNF](vnf.scad) using metaballs, or from
-//   general isosurfaces. Also provided are modules and functions to create [regions](regions.scad)
-//   (lists of polygon paths) for 2D metaballs and 2D contours.
+//   general isosurfaces. This file also provides modules and functions to create 2d metaballs and
+//   contours, where the output is a [region](regions.scad).
 //   .
-//   The point list in the generated VNF structure contains many duplicated points. This is normally not a
-//   problem for rendering the shape, but machine roundoff differences may result in Manifold issuing
-//   warnings when doing the final render, causing rendering to abort if you have enabled the "stop on
-//   first warning" setting. You can prevent this by passing the VNF through {{vnf_quantize()}} using a
-//   quantization of 1e-7, or you can pass the VNF structure into {{vnf_merge_points()}}, which also
+//   For isosurfaces and 3D metaballs, the point list in the generated VNF structure contains many duplicated
+//   points. This is normally not a problem for rendering the shape, but machine roundoff differences may
+//   result in Manifold issuing warnings when doing the final render, causing rendering to abort if you have
+//   enabled the "stop on first warning" setting. You can prevent this by passing the VNF through {{vnf_quantize()}}
+//   using a quantization of 1e-7, or you can pass the VNF structure into {{vnf_merge_points()}}, which also
 //   removes the duplicates. Additionally, flat surfaces (often resulting from clipping by the bounding
 //   box) are triangulated at the voxel size resolution, and these can be unified into a single face by
 //   passing the vnf structure to {{vnf_unify_faces()}}. These steps can be computationally expensive
@@ -1065,7 +1065,7 @@ function _bbox_sides(pc, pixsize, bbox) = let(
 ];
 
 
-function _contour_pixels(pixsize, bbox, fieldarray, fieldfunc, pixcenters, isovalue) = let(
+function _contour_pixels(pixsize, bbox, fieldarray, fieldfunc, pixcenters, isovalue, closed=true) = let(
     // get field intensities
     hp = 0.5*pixsize,
     field = is_def(fieldarray)
@@ -1101,7 +1101,7 @@ function _contour_pixels(pixsize, bbox, fieldarray, fieldfunc, pixcenters, isova
                 ],
                 pixcoord = [x,y],
                 pixfound_isoval = (min(pf) <= isovalue && isovalue <= max(pf)),
-                psides = _bbox_sides(pixcoord, pixsize, bbox),
+                psides = closed ? _bbox_sides(pixcoord, pixsize, bbox) : [],
                 pixfound_outer = len(psides)==0 ? false
                 : let(
                     ps = flatten([for(i=psides) _MTEdgeVertexIndices[i]]),
@@ -1645,6 +1645,7 @@ function debug_tetra(r) = let(size=r/norm([1,1,1])) [
 
 // Section: Metaballs (3D and 2D)
 //   ![Metaball animation](https://raw.githubusercontent.com/BelfrySCAD/BOSL2/master/images/metaball_demo.gif)
+//   ![Metaball animation](https://raw.githubusercontent.com/BelfrySCAD/BOSL2/master/images/metaball_demo2d.gif)
 //   .
 //   [Metaballs](https://en.wikipedia.org/wiki/Metaballs), also known as "blobby objects",
 //   can produce smoothly varying blobs and organic forms. You create metaballs by placing metaball
@@ -1653,8 +1654,111 @@ function debug_tetra(r) = let(size=r/norm([1,1,1])) [
 //   and melding together. The closer the objects are, the more they blend and meld.
 //   .
 //   The `metaballs()` module and function produces scenes of 3D metaballs. The `metaballs2d()` module and
-//   function produces scenes of 2D metaballs.
-
+//   function produces scenes of 2D metaballs. The metaball specification method, tranformations, bounding box,
+//   and other parameters are used the say way in 3D and 2D, but in 2D, pixels replace voxels. This
+//   introductory section describes features common to both 3D and 2D cases.
+//   .
+// Subsection: Common parameters for 3D and 2D metaballs
+//   **Parameter: `spec`**
+//   .
+//   The simplest metaball specification is a 1D list of alternating transformation matrices and
+//   metaball functions: `[trans0, func0, trans1, func1, ... ]`, passed as the `spec` parameter.
+//   Each transformation matrix you supply can be constructed using the usual transformation commands
+//   such as {{up()}}, {{right()}}, {{back()}}, {{move()}}, {{scale()}}, {{rot()}} and so on. You can
+//   multiply the transformations together, similar to how the transformations can be applied
+//   to regular objects in OpenSCAD. For example, to transform an object in regular OpenSCAD you
+//   might write `up(5) zrot(45) scale(4)`. You would provide that transformation
+//   as the transformation matrix `up(5) * zrot(45) * scale(4)`. You can use
+//   scaling to produce an ellipsoid from a sphere, and you can even use {{skew()}} if desired. 
+//   When no transformation is needed, give `IDENT` as the transformation.
+//   .
+//   The `spec` parameter is flexible. It doesn't have to be just a list of alternating transformation
+//   matrices and metaball functions. It can also be a list of alternating transforms and *other specs*,
+//   as `[trans0, spec0, trans1, spec1, ...]`, in which `spec0`, `spec1`, etc. can be one of:
+//   * A built-in metaball function name as described below, such as `mb_sphere(r=10)`.
+//   * A function literal accepting a 3-vector representing a point in space relative to the metaball's center.
+//   * An array containing a function literal and a debug VNF, as `[custom_func, [sign, vnf]]`, where `sign` is the sign of the metaball and `vnf` is the VNF to show in the debug view when `debug=true` is set.
+//   * Another spec array, for nesting metaball specs together.
+//   .
+//   Nested metaball specs allow for complicated assemblies in which you can arrange components in a logical
+//   way, or repeat a structure with different transformation matrices. That is,
+//   instead of specifying a transform and function, you specify a transform and then another metaball
+//   specification. For example, you could set `finger=[t0,f0,t1,f1,t2,f2]` and then set
+//   `hand=[u0,finger,u1,finger,...]` and then invoke `metaballs()` with `spec=[s0, hand]`. In effect, any
+//   metaball specification array can be treated as a single metaball in another specification array.
+//   This is a powerful technique that lets you make groups of metaballs that you can use as individual
+//   metaballs in other groups, and can make your code compact and simpler to understand. Keep in mind that
+//   nested components aren't independent; they still interact with all other components. See Example 24.
+//   .
+//   **Parameters: `bounding_box` and grid units**
+//   .
+//   The metaballs are evaluated over a bounding box. The `bounding_box` parameter can be specified by
+//   its minimum and maximum corners: `[[xmin,ymin,zmin],[xmax,ymax,zmax]]` in 3D, or
+//   `[[xmin,ymin],[xmax,ymax]]` in 2D. The bounding box can also be specified as a scalar of a cube (in 3D)
+//   or square (in 2D) centered on the origin. The contributions from **all**  metaballs, even those outside
+//   the box, are evaluated over the bounding box.
+//   .
+//   This bounding box is divided into grid units, specified as `voxel_size` in 3D or `pixel_size` in 2D,
+//   either of which can also be a scalar or a vector size.
+//   Alternately, you can set the grid count (`voxel_count` or `pixel_count`) to fit approximately the
+//   specified number of grid units into the bounding box.
+//   .
+//   Objects in the scene having any dimension smaller than the grid spacing may not
+//   be displayed, so if objects seem to be missing, try making the grid units smaller or the grid count
+//   larger. By default, if the voxel size or pixel size doesn't exactly divide your specified bounding box,
+//   then the bounding box is enlarged to contain whole grid units, and centered on your requested box.
+//   Alternatively, you may set `exact_bounds=true` to cause the grid units to adjust in size to fit instead,
+//   resulting in non-square grid units. Either way, if the bounding box clips a metaball and `closed=true`
+//   (the default), the object is closed at the intersection. Setting `closed=false` causes the object to end
+//   at the bounding box. In 3D, this results in a non-manifold shape with holes, exposing the inside of the
+//   object. In 2D, this results in an open-ended contour path with abiguity in how the path might be closed. 
+//   .
+//   For metaballs with flat surfaces or sides, avoid letting any side of the bounding box coincide with one
+//   of these flat surfaces or sides, otherwise unpredictable triangulation around the edge may result.
+//   .
+//   **Parameter: `isovalue`**
+//   .
+//   The `isovalue` parameter applies globally to **all** your metaballs and changes the appearance of your
+//   entire metaball object, possibly dramatically. It defaults to 1 and you don't usually need to change
+//   it. If you increase the isovalue, then all the objects in your model shrink, causing some melded
+//   objects to separate. If you decrease it, each metaball grows and melds more with others.
+//   .
+// Subsection: Metaballs debug view
+//   The module form of `metaballs()` and `metaballs2d()` can take a `debug` argument. When you set
+//   `debug=true`, the scene is rendered as a transparency (in 3D) or outline (in 2D) with the primitive
+//   metaball shapes shown inside, colored blue for positive, orange for negative, or gray for custom
+//   metaballs with no sign specified. These shapes are displayed at the sizes specified by the dimensional
+//   parameters in the corresponding metaball functions, regardless of isovalue. Setting `hide_debug=true` in
+//   individual metaball functions hides primitive shape from the debug view. Regardless the `debug` setting,
+//   child modules can access the metaball geometry via `$metaball_vnf` in 3D, or `$metaball_region` in 2D.
+//   .
+//   User-defined metaball functions are displayed by default as gray tetrahedrons (3D) or triangles (2D)
+//   with a corner radius of 5, unless you also designate a shape for your custom function, as described
+//   below in the documentation for {{metaballs()}} and {{metaballs2d()}}.
+//   .
+// Subsection: Metaballs run time
+//   The size of the grid units and size of the bounding box affects the run time, which can be long.
+//   Smaller grid units produce a finer, smoother result at the expense of execution time. Larger grid units
+//   shorten execution time.
+//   The affect on run time is most evident for 3D metaballs, less so for 2D metaballs.
+//   .
+//   For example, in 3D,
+//   a voxel size of 1 with a bounding box volume of 200×200×200 may be slow because it requires the
+//   calculation and storage of 8,000,000 function values, and more processing and memory to generate
+//   the triangulated mesh.  On the other hand, a voxel size of 5 over a 100×100×100 bounding box
+//   requires only 8,000 function values and a modest computation time. A good rule is to keep the number
+//   of voxels below 10,000 for preview, and adjust the voxel size smaller for final rendering. If you don't
+//   specify `voxel_size` or `voxel_count`, then a default count of 10,000 voxels is used,
+//   which should be reasonable for initial preview.
+//   .
+//   In 2D, If you don't specify `pixel_size` or `pixel_count`, then a default count of 1024 voxels is used,
+//   which is reasonable for initial preview. You may find, however, that 2D metaballs are reasonably fast
+//   even at finer resolution.
+//   .
+//   Because a bounding box that is too large wastes time
+//   computing function values that are not needed, you can also set the parameter `show_stats=true` to get
+//   the actual bounds of the voxels intersected by the surface. With this information, you may be able to
+//   decrease run time, or keep the same run time but increase the resolution. 
 
 // Function&Module: metaballs()
 // Synopsis: Creates a group of 3D metaballs (smoothly connected blobs).
@@ -1666,38 +1770,11 @@ function debug_tetra(r) = let(size=r/norm([1,1,1])) [
 // Usage: As a function
 //   vnf = metaballs(spec, bounding_box, voxel_size, [isovalue=], [closed=], [exact_bounds=], [convexity=], [show_stats=]);
 // Description:
-//   The simplest metaball specification is a 1D list of alternating transformation matrices and
-//   metaball functions: `[trans0, func0, trans1, func1, ... ]`, passed as the `spec` parameter.
-//   Each transformation matrix you supply can be constructed using the usual transformation commands
-//   such as {{up()}}, {{right()}}, {{back()}}, {{move()}}, {{scale()}}, {{rot()}} and so on. You can
-//   multiply the transformations together, similar to how the transformations can be applied
-//   to regular objects in OpenSCAD. For example, to transform an object in regular OpenSCAD you
-//   might write `up(5) xrot(25) zrot(45) scale(4)`. You would provide that transformation
-//   as the transformation matrix `up(5) * xrot(25) * zrot(45) * scale(4)`. You can use
-//   scaling to produce an ellipsoid from a sphere, and you can even use {{skew()}} if desired. 
-//   When no transformation is needed, give `IDENT` as the transformation.
+//   Computes a [VNF structure](vnf.scad) of a 3D metaball scene within a specified bounding box.
 //   .
-//   The metaballs are evaluated over a bounding box. The `bounding_box` parameter can be specified by
-//   its minimum and maximum corners `[[xmin,ymin,zmin],[xmax,ymax,zmax]]`,
-//   or specified as a scalar size of a cube centered on the origin. The contributions from **all**
-//   metaballs, even those outside the box, are evaluated over the bounding box. This bounding box is
-//   divided into voxels of the specified `voxel_size`, which can also be a scalar cube or a vector size.
-//   Alternately, you can set `voxel_count` to fit approximately the specified number of boxels into the
-//   bounding box.
-//   .
-//   Smaller voxels produce a finer, smoother result at the expense of execution time. Larger voxels
-//   shorten execution time. Objects in the scene having any dimension smaller than the voxel may not
-//   be displayed, so if objects seem to be missing, try making `voxel_size` smaller or `voxel_count`
-//   larger. By default, if the voxel size doesn't exactly divide your specified bounding box, then the
-//   bounding box is enlarged to  contain whole voxels, and centered on your requested box. Alternatively,
-//   you may set `exact_bounds=true` to cause the voxels to adjust in size to fit instead. Either way, if
-//   the bounding box clips a metaball and `closed=true` (the default), the object is closed at the
-//   intersection surface. Setting `closed=false` causes the [VNF](vnf.scad) faces to end at the bounding
-//   box, resulting in a non-manifold shape with holes, exposing the inside of the object.
-//   .
-//   For metaballs with flat surfaces (the ends of `mb_cyl()`, and `mb_cuboid()` with `squareness=1`),
-//   avoid letting any side of the bounding box coincide with one of these flat surfaces, otherwise
-//   unpredictable triangulation around the edge may result.
+//   The [subsection on parameters](#subsection-common-parameters-for-3d-and-2d-metaballs) above describes in
+//   detail how the primary parameters work for metaballs(). The `spec` parameter lets you define the
+//   metaballs in your scene, including their shape, position, and orientation.
 //   .
 //   You can create metaballs in a variety of standard shapes using the predefined functions
 //   listed below. If you wish, you can also create custom metaball shapes using your own functions
@@ -1732,29 +1809,6 @@ function debug_tetra(r) = let(size=r/norm([1,1,1])) [
 //   Negative metaballs are never directly visible; only their effects are visible. The `influence`
 //   argument may also behave in ways you don't expect with a negative metaball. See Examples 16 and 17.
 //   .
-//   The `spec` parameter is flexible. It doesn't have to be just a list of alternating transformation
-//   matrices and metaball functions. It can also be a list of alternating transforms and *other specs*,
-//   as `[trans0, spec0, trans1, spec1, ...]`, in which `spec0`, `spec1`, etc. can be one of:
-//   * A built-in metaball function name as described below, such as `mb_sphere(r=10)`.
-//   * A function literal accepting a 3-vector representing a point in space relative to the metaball's center.
-//   * An array containing a function literal and a debug VNF, as `[custom_func, [sign, vnf]]`, where `sign` is the sign of the metaball and `vnf` is the VNF to show in the debug view when `debug=true` is set.
-//   * Another spec array, for nesting metaball specs together.
-//   .
-//   Nested metaball specs allow for complicated assemblies in which you can arrange components in a logical
-//   way, or repeat a structure with different transformation matrices. That is,
-//   instead of specifying a transform and function, you specify a transform and then another metaball
-//   specification. For example, you could set `finger=[t0,f0,t1,f1,t2,f2]` and then set
-//   `hand=[u0,finger,u1,finger,...]` and then invoke `metaballs()` with `spec=[s0, hand]`. In effect, any
-//   metaball specification array can be treated as a single metaball in another specification array.
-//   This is a powerful technique that lets you make groups of metaballs that you can use as individual
-//   metaballs in other groups, and can make your code compact and simpler to understand. Keep in mind that
-//   nested components aren't independent; they still interact with all other components. See Example 24.
-//   .
-//   The isovalue parameter applies globally to **all** your metaballs and changes the appearance of your
-//   entire metaball object, possibly dramatically. It defaults to 1 and you don't usually need to change
-//   it. If you increase the isovalue, then all the objects in your model shrink, causing some melded
-//   objects to separate. If you decrease it, each metaball grows and melds more with others.
-//   .
 //   ***Built-in metaball functions***
 //   .
 //   Several metaballs are defined for you to use in your models. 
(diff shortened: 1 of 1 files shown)


amatulic:
I may not be online for a day due to travel.

To do:
* Check through the docs. They are now organized into two sections: metaballs and isosurfaces/contours. I moved a few words from the modules into the intro of each section.
* Test whether anchoring and attachments are working.
* Figure out what happens to anchoring and attachments in debug view. stroke() doesn't seem to support anchoring. Maybe it would be best to create an outline of the shape by subtracting a small negative offset.
* It would be nice to create an animation for 2D. I'm thinking of the various shapes rotating and bouncing off the walls of a box. With a square box, the path of each object would be repeatable.
* Come up with 2D examples that actually relate to 3D design. @RAMilewski do you have any ideas? 2D features would be useful for people with laser cutters instead of 3D printers, or people wanting to make signs on their 3D printers. Possibilities:
  * Simulated wood grain pattern: randomly distribute random-size circle metaballs, extreme-stretch it in one direction. Increase or decrease size of balls for different concentric patterns.
  * Recognizable animal for a 2D sign. A duck?
  * A hand for a warning sign.
  * A surface texture for rotate_sweep, to make decorative pots and planters. Water waves? I don't know how this would work.
  * 2D contours could possibly be used in a loft, although for what purpose I don't know.

adrianVmariano:
The stroke module doesn't support anchoring because it doesn't know the dimensions of its object.

It may be the case that your cleanest solution to support anchoring is to go back to using attachable() directly, because you can give it the region and it won't care how you instantiate it in geometry.   Easiest way to get that right is probably to look at how the region() module calls attachable and just do that.  

`metaballs2d([IDENT, mb_circle(10)], bounding_box=12);` produces an error about unknown variable.  

While fiddling around I expected `mb_stadium([w,h])` to work---it seems like a natural alternative to `mb_rect()`.   This seems more natural and fits with other shapes better than a separate width and radius, I think.  

" Also provided are modules and functions to create [regions](regions.scad)
//   (lists of polygon paths) for 2D metaballs and 2D contours are also supported."

Wording awkward.  "Also provided are modules and functions to create 2d metaballs and contours, where the output is a region."

I'm uncertain if the approach of referencing 3d metaballs is the right one to use for documenting this.  A user may land directly on metaballs2d and not have seen metaballs3d.   But I suppose you do have a link.   But you do repeat some material, and it's arguably the least important material (on custom metaballs).  Need to change 3-vector to 2-vector there.    If you want to avoid repeating material the best way might be to create a generic metaballs description separate from either module that they both link to and then more terse but complete arg explanations in each 2d and 3d case.  

I would say "in 2D, pixels replace voxels" rather than "we refer to".   The text "e.g. sphere accepts ..." should change to circle.  

The trapezoid module/function takes a bunch of args you don't accept, so I wouldn't say "same dimensional arguments".   It might be nice to accept the angle arg (which would have to be scalar since it has to be symmetric).    Last phrase "which ARE less flexible".  

stadium is described as a cylinder, which is wrong.  As noted above consider accepting [w,h].   It doesn't "resemble" the convex hull of two circles....it is that.   I'm not sure that description is helpful when the circles/spheres are the same size---maybe you wrote it when you thought the cone version was going to work?    Same is true for mb_capsule, actually about "resembling".   The stadium function does NOT use similar args to cyl().  

Connector isn't really a "rod" and certainly not of radius anything.  Connecting rectangle with rounded ends?   Again ditch "resemble".   Do the interpolation artifacts also occur in 2d?   I tried to create some and wasn't successful...though it seems like the mechanism that makes them should still be at work, so they should be happening.  

Note that mb_ring has a more limited parameter set than ring?

px_centers is a bad arg name.  Brainstorming: refine=true, use_centers=true, augment=true

show_box is said to use a "transparent thin rectangle".  Delete the word "thin".  (What is a thin rectangle?)  Description of debug is wrong (full of 3d references).   There is no "orient" in 2d.    And anchor should translate to (0,0).    Hmmm.  This questionable description of anchor translating to (0,0,0) appears all the time in shapes2d.   It's also a bit unclear that "about the z axis" should be stated for spin.  

The contour docs say f(x,y,z) in first sentence of 2nd paragraph.  

Comparing run time to isosurface doesn't seem useful.  Better to assume the user has never run isosurface when they read this.   One thing that may more strongly affect run time in 2d is the number of path segments, because the path assembly process is slow.  Have you observed that?  So like an empty region with lots of pixels is going to be maybe much faster than a region full of contours.   

Same complaint about px_centers.  My best alternative so far is refine=true

Default for smoothing sounds like it's incorrectly stated.  It should be "Default: 2 if px_centers=true and 0 otherwise."   Remember, avoid talking about "undef" in the docs unless it's the return value.  And definitely "undef" cannot be a default.  
 
Note that instead of iterating a length 2 filter you might just apply one filter of the desired length one time, though I guess that would be more complicated to implement.  (It would be more flexible since the filter's form would be more free.)   So basically your single filter is [1,1]/2 and iterating twice is [1,2,1]/4, and 3x is [1,3,3,1]/8.   Are those the best choices for filters of lengths 2, 3 and 4?   Note also that the convolve() function should perform this operation for you.  Describing smoothing by kernel length may be a bit more informative than iteration count.  It directly tells you how far the influence extends from the smoothing---can path point i affect path point j?

amatulic:
> probably to look at how the region() module calls attachable and just do that.

When `debug=true`, it now uses attachable() the same way region() does.

> `metaballs2d([IDENT, mb_circle(10)], bounding_box=12);` produces an error about unknown variable.

Fixed.
 
> While fiddling around I expected `mb_stadium([w,h])` to work---it seems like a natural alternative to `mb_rect()`.

Size arguments that can be a vector can also be a scalar, which wouldn't really make sense here (you just get a sphere). I did make it so that it takes the arguments h, w (not a vector), consistent with the other metaball functions.

I changed the documentation as you suggested. However.

> I'm uncertain if the approach of referencing 3d metaballs is the right one to use for documenting this.

I'm busy reorganizing the documentation to discuss everything common to both 3D and 2D metaballs in one place. This is most of the wordy stuff.

> The trapezoid module/function takes a bunch of args you don't accept, so I wouldn't say "same dimensional arguments". It might be nice to accept the angle arg

Added angle arg, using the trapezoid private function to get h, w1, and w2 from any combination of 3 of h, w1, w2, and ang.

> Note that mb_ring has a more limited parameter set than ring?

By design. ring() parameters seem overly complex for metaballs.

> px_centers is a bad arg name. Brainstorming: refine=true, use_centers=true, augment=true

Changed to `use_centers`. It doesn't actualy refine or agument, in fact it make things look more jagged due to following, in effect, triangular facet surfaces of a 3D object. Smoothing is needed to correct the jaggies.

> Comparing run time to isosurface doesn't seem useful. Better to assume the user has never run isosurface when they read this. One thing that may more strongly affect run time in 2d is the number of path segments, because the path assembly process is slow. Have you observed that?

I haven't really observed any significant slowdown due to path assembly. metaballs2d() runs pretty fast compared to 3D. Also, 2D contours look reasonable using bigger pixels than one would need in voxels for 3D.

> Note that instead of iterating a length 2 filter you might just apply one filter of the desired length one time, though I guess that would be more complicated to implement. ... So basically your single filter is [1,1]/2 and iterating twice is [1,2,1]/4, and 3x is [1,3,3,1]/8. Are those the best choices for filters of lengths 2, 3 and 4? Note also that the convolve() function should perform this operation for you.

The user also doesn't need to know the details of the polygon data. Yes, those are just binomial coefficients, and I could use convolve, although I am treating things differently if closed=true versus closed=false, which is somewhat simpler if I just iterate the two-point filter n times. I am uncertain, however, what the usefulness of the `closed` parameter is for a 2D contour.

adrianVmariano:
Just because mb_stadium is not useful with a scalar argument does not mean that it makes sense to create an inconsistent interface.  Every object with multiple dimensions accepts a vector for dimensions, so having two parameters is inconsistent.   A user trying to guess the parameters will try [w,h] like I did.  If you find it highly objectionable that a scalar makes a sphere you can not accept that case---but do you ban users giving w=h on input?  

When I said "Note that mb_ring has a more limited parameter set than ring?" I was suggesting that in the docs, you add a note to this effect.  I wasn't trying to draw your attention to it like it needed fixing.  

So is run time dominated by the marching squares, not the path assembly?   

I hadn't noticed that convolve lacks an option for circular convolution.  Probably should be added....  

What do you mean when you say "I am uncertain what the usefulness of the closed parameter is for a 2d contour"?

amatulic:
The intent was for the 2D metaballs to have a similar parameter set as the 3D metaballs. mb_stadium is the analogue to mb_disk and mb_capsule, which take arguments h and d like cylinder, which is why mb_stadium takes h and d also (or h and w). Because there is no BOSL2 equivalent to stadium, I felt this was appropriate, but I don't have a strong objection to folding those arguments into a vector other than it's inconsistent with the 3D counterparts.

It occurs to me that for 2D, if w>h, that could be the same as a stadium rotated 90°.

I haven't done extensive run time testing because I haven't yet been annoyed by slow run times. What would be worst case test condition, 1024 segments assembled into 1 path, or 1024 segments that are already independent paths?

About the usefulness of the `closed` parameter, I meant that I can't think of a use case for changing it from the default closed=true. A contour is *always* closed, even if clipped by the bounding box, whatever's beyond is closed. I suppose for an infinite function like a gyroid, you'd get endless unclosed contour lines, but then again, what is the use case for that?

adrianVmariano:
So rect() is the 2d equivalent of the 3d cyl() but rect() does not have the same or analogous parameters.  I think for stadium a vector pair is best. 

Seems reasonable to make the w>h case work that way.  

I think that worst case would be 1024 segments that are independent paths.  

You're saying the closed= option isn't useful for metaballs or contours?  I think it could definitely be useful.  Someone may want an open contour to continue it other ways or use somehow in  a downstream computation.   However, there's the little detail that a list of open paths is no longer a region.  So you may want to describe it as "list of paths" in the closed=false case.

amatulic:
Sure I'll leave in the `closed` option. But all closed=true does is create more segments along a side of the bounding box if any contour was clipped. If there are no clipped contours, then the region returned would be identical regardless of whether closed is true or false. In this case "closed" doesn't mean "closed paths" in the context of regions, it means "connect path endpoints that touch the bounding box" -- and that doesn't mean an individual path would become closed, it could mean that path gets connected to some other path that also touches the bounding box, depending on the function values along the box edge.

Maybe the term "closed" is being overloaded here. It means something similar in 3D. It doesn't mean what it means elsehwere in BOSL2.

adrianVmariano:
I tried this case

```
fdata=contour(function(x,y) sin(150*sqrt(x^2+y^2)), bounding_box=[[0,-11],[11,11]],pixel_size=.5, isovalue=.7, closed=false);
rainbow(fdata) stroke($item, width=.05);
```

and found the behavior confusing.  It seemed the same regardless of closed.  It appeared that the segments along the bounding box were included even if closed=false.  I expect it to return paths that have holes along the edge of the bounding box when closed=false.  It's also a wee bit surprising that closed=true produces contours that are rounded at the bounding box.  That's due to the smoothing, presumably.  But it might be worth calling out this specific behavior of the smoothing and "if you want sharp corners at the bounding box turn it off".

adrianVmariano:
If closed=false then is it not the case that you do not get the clipped edges?  That would mean that the contour would be wrong if the clipped edge isn't collinear, e.g. if it's a corner of the bounding box.   It also means you wouldn't know in the case of multiple contours how to connect them.  

If closed=true you get all of those edges along the bounding box and then when you run path_merge_collinear at the end they get converted into one long segment.  But if you have several components this will produce a different result that isn't necessarily easy to make from the above, it seems.

amatulic:
A test for `closed` was missing from the 2D pixel-finding function, which is present in the 3D voxel-finding function. It was doing closed=true all the time.

If closed=false you get contours ending at the bounding box edges. If closed=true you get the ends at the bounding box edges connected to other ends at the bounding box edges.

adrianVmariano:
You're missing the point about region.  A region is defined as a set of non-intersecting **closed** curves.  A list of non-closed curves is not a region, so it doesn't make sense to pass such an object to region().  It won't be an error because the region() module accepts non-regions, and in some cases may produce a result that doesn't look different, but if paths terminate on different edges of the box you're likely to get weird effects with diagonal lines across the shape and parts of the result xored together.  My point is that you can only pass the result to region if it includes the clipped edge segments, which would be the `closed=true` condition.  If `closed=false` and you don't compute those clipped segments, you have a result which should not be passed to `region()`.

amatulic:
I see. I'm wondering if it would be useful to return a data structure rather than a region or list of paths. It would include the list of paths, and another array full of booleans corresponding to each path saying whether the path is closed or intersects the bounding box. Maybe if closed=false I can echo a notice to the console and draw filled polygons for the closed paths and use stroke for the unclosed paths. This might complicate anchoring.

Can you figure out the docsgen error? It seems to have something to do with section names, but I am not sure what. Maybe it's the way I link to sections. I don't know.

adrianVmariano:
I'm inclined to say that if closed=false, it's too hard to decide on the right thing to do.  You could run stroke, I suppose.  But the user needs the width parameter, and what if they want other stroke features?  It feels like a mistake to go there.  Such a user can use the function form and then call stroke.   I'm still thinking that the best behavior is to only support `closed=true` for the module forms.  

The problem with returning a more complicated data structure is that it's mostly going to be a nuisance except for the 0.1% of users that find it useful for something.   Everybody else is going to be doing stroke(returned_value[0]) all the time.  Here's an idea.  When closed=false, leave paths that intersect the boundary as they are but explicitly close paths that do not intersect the boundary by duplicating their starting vertex (I think list_wrap is the function that does that).  This is inconsistent with general BOSL2 conventions, but things like stroke will work.  And a user can strip those extra points if necessary.   But also it's easy to test if a component was closed or not by checking of `thing[0]==last(thing)`.

adrianVmariano:
Another possible way to return a mixture of open and closed paths would be to return two lists, one list of the open paths that hit the bounding box and a second list of the closed paths that do not.  This would be less transparent for passing to stroke and other basic uses, though.

amatulic:
That's a good idea to close the closed paths. I believe that can already be done by commenting out this line in _assemble_partial_paths_recur():
`: left[0] != right[0] ? concat(paths[left[0]], paths[right[0]])`

The soup of segments it gets initially includes all segments for a closed path with the beginning and end connected, but the final segment is dropped when the paths are assembled.

adrianVmariano:
No, don't do that because you only want to close the closed paths in the `closed=false` case so they can be distinguished from open paths.  I think passing a closed path to, for example offset() will fail (or do the wrong thing) so it's not ideal to do this.  When the user gives closed=true they should get a normal region where the paths do not duplicate endpoints, but are closed.

adrianVmariano:
In looking for things that could cause the error.  

I see that you have a "." on a line before each subsection.  I never did that.  (Don't know why it would be an issue....but you could remove those.) 

I did finally find your references.  They are almost certainly broken, though I don't know if they are the problem.  There are broken refs in utility.scad as well that are just..well...broken.   Based on other examples it looks like they need to look like:

`(isosurface.scad#section-the-full-heading-name)`

And I have no idea if you can have a reference to a name with parentheses in it.  Seems like that might confuse docsgen.  You just ignored parentheses in your reference.

amatulic:
I think I found the error: a blank line before a "Subsection" heading. I still need to do some more reorgaizaing of the docs.

The way I am detecting closed paths is any path that doesn't share a coordinate with one of the bounding box borders. There might be a rare case where one vertex touches the border (like if a pixel corner is exactly at the isovalue at the border). I thought it might be more reliable if the path assembly function closed them for me. It could do a check for whether `closed` is true or false.

adrianVmariano:
Well, if it's easier and more reliable to have path assembly close them, then you could do that and when closed=true you could unclose all the paths afterwords.   Or you could add an option to path assembly about whether to add the closing segment.

adrianVmariano:
Did you fix mb_disk as noted in #1596 ?

amatulic:
I put in a check so that list_unwrap() is done only if closed=true. I also fixed 1596.

If closed=false, the module draws the polygon if path[0]==last(path), and for unclosed paths it does stroke() using a width 1/4 the pixel size. I updated the documentation to explain that when closed=true you can expect a valid region returned with all closed paths, but if closed=false you get a mixture, and closed paths have equal end points.

adrianVmariano:
In looking at the error messages I think the problem for docsgen is that you have subsections in a description block.   You also have a subsection in a block by itself, which could be a problem.   You have really short subsections.  They get a kind of huge heading.  They also get listed in the contents list and linked.   It's not clear that it makes sense to have subsubsections the way you do, or rather, to have more than two levels.

adrianVmariano:
As noted before, I think that invoking stroke with constant width is probably not very useful except for debugging or visualization purposes.  A user trying to actually make something needs to control the width, at the very least.  Do I want a shell in my model that is 1/4 a pixel wide?  Probably not unless pixels are at least 4 units.  Will my result look good with pixels that big?

adrianVmariano:
I had a realization than the 2nd stage of path assembly that looks for loops assumes and treats the path as a closed path.   I'm not sure how easy it is to fix that, but it's potentially a serious problem because the closing segment it will add could cross other parts of the path and create bogus loops.

adrianVmariano:
This entry is a bit rambling....

```
f = function(x,y) sin(75*x)*cos(75*y);
contour(f, bounding_box=15, isovalue=.000,closed=true, pixel_size=.5);
```

Gave me:

image

and

```
contour(f, bounding_box=15, isovalue=-.03,closed=true, pixel_size=1);
```

gives one of those holes:

image

Isovalue of -0.1 gives even more extra holes.  Or extra lines?  Not sure which thing is extra.  

```
a=contour(f, bounding_box=15, isovalue=-.1,pixel_size=.2);
``` 

Some postprocessing is happening after path assembly that messes up the paths.  I printed out the input paths to path assembly and displayed them with a rainbow stroke and got this:

image

But if I display the result after processing it looks like this:

image

At a pixel size 0.75 I get this, which looks like it could possibly be related to the mystery holes?  Not sure, since I'm making it return data that isn't what you're expecting downstream.  But even if you close all those segments, it shouldn't be removing the connecting endpoints.  That is, I should have a sequence of contacting loops.  So a little puzzed here.  

image

Hmmm.  Actually I think smoothing may be causing the problem.  I disabled my hack and found that smoothing introduced the extra hole.  

```
a=contour(f, bounding_box=15, isovalue=-.1,pixel_size=.75,smoothing=0);
rainbow(a,shuffle=true) stroke($item,width=.02);
```

Gave me this:

image

Is that a bug in path assembly?  Why the short green segment?  

Smoothing turns it to this:

image

```
a=contour(f, bounding_box=[[-1,3],[-7.5,-6.1]], isovalue=-.1,pixel_size=.75,smoothing=1);
```

produced error:
```
[WARNING: len() parameter could not be converted: argument 0: expected string, found undefined (undef) in file isosurface.scad, line 1081](https://github.com/BelfrySCAD/BOSL2/pull/1081,C:/Users/adrian/Documents/OpenSCAD/isosurface.scad)
[TRACE: called by 'len' in file isosurface.scad, line 1081](https://github.com/BelfrySCAD/BOSL2/pull/1081,C:/Users/adrian/Documents/OpenSCAD/isosurface.scad)
[TRACE: called by '_contour_pixels' in file isosurface.scad, line 3619](https://github.com/BelfrySCAD/BOSL2/pull/3619,C:/Users/adrian/Documents/OpenSCAD/isosurface.scad)
[TRACE: called by 'contour' in file joiner11.scad, line 592](https://github.com/BelfrySCAD/BOSL2/pull/592,C:/Users/adrian/Documents/OpenSCAD/joiner11.scad)
[TRACE: assignment to "a" in file joiner11.scad, line 592](https://github.com/BelfrySCAD/BOSL2/pull/592,C:/Users/adrian/Documents/OpenSCAD/joiner11.scad)
```

Looks like the input to path assembly has reversed segments:

```
a=contour(f, bounding_box=[[-2+1.5,-7.5],[4.4-1.5-1.5,-4.5-1.5]], isovalue=-.1,pixel_size=.75,smoothing=1);
```

After path assembly but before loop removal or any further postprocessing I get this:

image

So the red segment and blue segment go in opposite directions, and hence don't get assembled.  That's the source of your loops.  Not sure if it's the source of **all** the weirdness.

adrianVmariano:
```
![image](https://github.com/user-attachments/assets/790515d7-8895-4263-9be7-b1bb2d2b4702)
```

with f as above if it matters producxes:

```
[WARNING: len() parameter could not be converted: argument 0: expected string, found undefined (undef) in file libraries/BOSL2/lists.scad, line 286](https://github.com/BelfrySCAD/BOSL2/pull/286,C:/Users/adrian/Documents/OpenSCAD/libraries/BOSL2/lists.scad)
[TRACE: called by 'len' in file libraries/BOSL2/lists.scad, line 286](https://github.com/BelfrySCAD/BOSL2/pull/286,C:/Users/adrian/Documents/OpenSCAD/libraries/BOSL2/lists.scad)
[TRACE: called by 'last' in file isosurface.scad, line 1179](https://github.com/BelfrySCAD/BOSL2/pull/1179,C:/Users/adrian/Documents/OpenSCAD/isosurface.scad)
[TRACE: called by 'norm' in file isosurface.scad, line 1179](https://github.com/BelfrySCAD/BOSL2/pull/1179,C:/Users/adrian/Documents/OpenSCAD/isosurface.scad)
[TRACE: called by '_assemble_partial_paths_recur' in file isosurface.scad, line 1192](https://github.com/BelfrySCAD/BOSL2/pull/1192,C:/Users/adrian/Documents/OpenSCAD/isosurface.scad)
[TRACE: called by '_assemble_partial_paths' in file isosurface.scad, line 3625](https://github.com/BelfrySCAD/BOSL2/pull/3625,C:/Users/adrian/Documents/OpenSCAD/isosurface.scad)
[TRACE: called by 'contour' in file isosurface.scad, line 3584](https://github.com/BelfrySCAD/BOSL2/pull/3584,C:/Users/adrian/Documents/OpenSCAD/isosurface.scad)
[TRACE: assignment to "pathlist" in file isosurface.scad, line 3584](https://github.com/BelfrySCAD/BOSL2/pull/3584,C:/Users/adrian/Documents/OpenSCAD/isosurface.scad)
[TRACE: called by 'contour' in file joiner11.scad, line 591](https://github.com/BelfrySCAD/BOSL2/pull/591,C:/Users/adrian/Documents/OpenSCAD/joiner11.scad)
Execution aborted

**Comment by amatulic** [comment]
I verified that all your examples above are now working after correcting the path direction of that one pixel ID. No holes, no error messages, everything is well behaved. That pixel tripped me up because I was thinking "clockwise paths" and the path was winding around clockwise inside the pixel -- but it should wind clockwise around regions higher than the isovalue, meaning that higher pixel corners should always be on the right of the path, and in that case it should go around the inside of the pixel counterclockwise.

OK, here are the approaches I thought of to identify closed and open paths:
* Have the assembly function make closed paths with duplicate endpoints
* Test whether any vertices are within epsilon of the bounding box
* Call the assembly function twice, once to get paths before introducing the clipping segments, and once again to include the clip segments to close all the paths - the second pass should be fast because all the paths would already be assembled except for the segements on the bounding box. It wouldn't even need to be called a second time if there are no bounding box segments. This would require two passes through the pixel array, though.

**Comment by adrianVmariano** [comment]
I think if you want path assembly to handle open paths you can do it by changing the loop finding code to look for that case with a `closed=false` operating mode.  Where it creates the path fragments you would need to check for "tails" on the ends of the path.  The tails would need to be checked to see if they are one path or two paths and then loop splitting could run on what's left, which would be closed, and then the tail portion added back.

**Comment by adrianVmariano** [comment]
Note that you don't test whether **vertices** are within epsilon of the bounding box, you test whether **edges** are within epsilon of the bounding box.  Is it the case that your code can produce clipped edges that are not exactly on the bounding box?  I would have thought that since you clip them to the box they would be exactly on it.  

I'm not sure I understand your list of proposed methods.  Right now path assembly **requires** that all paths be closed.  It will potentially produce bogus output if you feed it non-closed paths.  So what are the options?  

* In the closed=true state return a region as produced by path assembly.  In the closed=false state, apply list_wrap to all path components and then detect and remove clipped edges by comparing to the bounding box.    This seems straight forward to implement.  Are there downsides? 
* Change the path assembly to support open paths as well as closed paths and give it two operating modes, one that works on closed paths and one that works on open paths.  The main downside here is that this seems like a harder thing to implement and debug and makes the path assembly more complicated.  The path assembly code may have broader utility.    Does this method have an upside to warrant that extra effort?  
* Was running path assembly twice meant to allow averaging to be applied only to non-clipped edges?   Why does this requires two passes through the pixel array?  That seems undesirable and we expect the closed=true case to be the normal case, so it puts more computation on the more common case if you need a second pass to do the clipping.  This approach would also require that path assembly be modified to work on open paths.  It seems like you could confine averaging to non-clipped edges if you flag the bounding edges and have the averaging skip the flagged sections where you just hold them fixed.   Or you can just skip sections as you go if they are on the bounding box.  (If it's not clear, your averaging is taking two points adjacent points and averaging them together to get a new point.  So if the point is part of a bounding edge you just return the unaltered point instead of computing the average.)  

Was the error that I got when I didn't give a pixel size somehow connected to the direction reversal bug?

**Comment by adrianVmariano** [comment]
In the closed and unclosed paths section I thought it was weird to introduce a parameter and they say it's not supported and then say what the default value of this nonexistent parameter is.   I suggest you don't mention the module version, since this is not a feature of the module, and say just:

The functional form of metaballs2d() support a `closed` parameter.  When `closed=true` (the default) and a polygon is clipped by the bounding box, the bounding box edges will be included in the polygon.  The resulting path list is a valid region with no duplicated vertices in any path.  

When `closed=false` the paths end at the bounding box, which means that the list of paths may include a mixture of closed and open paths.  Regardless of whether any of the output paths are open, all closed paths will have identical first and last points so that closed and open paths can be distinguished.  You can use {{are_ends_equal()}} to determine if a path is closed.  A path list that includes open paths is not a region, since regions are lists of **closed** polygons.  Duplicating the ends of closed paths can cause problems for some functions such as {{offset()}} which will complain about repeated points; to deal with this problem you can pass the closed components to {{list_unwrap()}} to remove the extra endpoint.  

--

Note that duplicating points in a path does not make a region invalid.  It breaks a convention and some things like offset (and I'm not sure what else) don't work.  The reason for not just eliminating points in offset so that things work (which I suspect is your immediate reaction) is that this changes the point count, and if you're trying to make lists with N points to link them together that could be confusing, like you start with N points and feed it to offset with the same_length=true parameter and are mystified that only N-1 points come out...which you don't realize at first but only later downstream when you get an error from vnf_vertex_array....

I'm not seeing an explanation for what metaballs2d the module does, so it might be appropriate to say something about its behavior, but that shouldn't require reference to nonexistent parameters, just "The metaballs module does X".   And it shouldn't be in a subsection with a title of a parameter that doesn't exist for the module.  

In looking at how things are organized now, I see that you've got a section called "Metaballs (3D and 2D)".  It might be titled just "Metaballs" or something else.  Since there is no section describing 4D metaballs or 1D metaballs giving the dimension doesn't really add.  

One question to think about is what information to relegate to that intro and what to repeat in the descriptions of the individual functions.  I think that the information on custom metaballs functions is the LEAST important content present and that should definitely be relegated to the intro section and not repeated.    It is also the content that goes into the most technical detail and hence in some respects, the sort of thing that **should** be in that introductory section.   If you just change 3-vector and 2-vector into "vector" the text applies to both cases.  

" The `spec` parameter is described in more
//   detail here."

The last word "here" should be "there".  

I might write the first line more compactly as "See [metaball parameters] for details on the primary parameters common to metaballs() and metaballs2d()."  If you leave it alone I'd not call it a "subsection" since it isn't one and I'd strike "above" since a user using a hyperlink doesn't care.  It may make sense to at least describe the simplest form of spec in the description before passing the user off with the link.

**Comment by amatulic** [comment]
I don't understand why the path assembly requires that all paths be closed. Feeding the path assembly function non-closed paths doesn't seem to be a problem. Here's what may be a worst-case, with four paths, all of them clipped, and all of them including a bounding box corner.
````
include <BOSL2/std.scad>
include <BOSL2/isosurface.scad>

field =[
    [2,2,2,2,0,2,3,4,4],
    [2,0,1,2,0,2,2,3,4],
    [2,1,0,1,0,2,2,2,2],
    [2,2,1,0,0,2,2,2,2],
    [0,0,0,0,0,0,0,0,0],
    [2,2,2,2,0,0,1,2,2],
    [2,2,2,2,0,1,0,1,2],
    [2,2,2,2,0,2,1,0,2],
    [2,2,2,2,0,2,2,2,2]
];
isoval=0.7;
pixsize = 5;
c = contour(field, isoval, pixel_size=pixsize,
        closed=true,
        use_centers=true);
zrot(-90) for(p=c) stroke(p, width=0.3);
````
Here it is for closed=true:
image

Here it is for closed=false:
image

It looks perfect to me.

**Comment by adrianVmariano** [comment]
I believe that the path assembly is written under the assumption that paths are closed---specifically the part that does loop removal.  (The loop removal code is part of the support for operations like intersection on regions and regions are always made from closed paths.)  Since none of your examples contain loops you didn't actually test the potentially problematic case.   It occurs to me that since intersections are found by a point search instead of with actual line intersections, the most obvious failure mode I was thinking about won't occur, so I'm not sure if there's a problem or not.  (Recall that I didn't write that loop removing code.)    I think that it's necessary to study the code to figure out what happens with open paths if you want to make sure that no failure modes exist.  

Does two passes add significantly to the time compared to a one pass approach?  You have saved the locations of the boxes where the curve is for the 2nd pass?  

You should probably not permit more ways to specify the bounding box, but if my box is bogus I should get an error message to that effect, not a cryptic failure in the code.   That is, there should be an all_positive(box[1]-box[0]) assertion after a is_matrix(box,2,2) assertion.

**Comment by revarbat** [comment]
If I recall correctly, open paths never find a completing segment, and are added to the returned path list verbatim. You can distinguish them from closed complete paths by checking that the first and last vertex are not equal.

**Comment by amatulic** [comment]
@revarbat - Here's the code we're talking about in isosurface.scad. Adrian wrote it based on other code in BOSL2. For all paths returned, the first and last values are not equal, but it doesn't seem to care whether paths are open or closed, it works correctly.

````
function _assemble_partial_paths(paths,closed=false,eps=EPSILON) =
    let(
        pathlist = _assemble_partial_paths_recur(paths),
        splitpaths = 
            [for(path=pathlist) each
               let(
                   searchlist = vector_search(path,eps,path),
                   duplist = [for(i=idx(searchlist)) if (len(searchlist[i])>1) i]
               )
               duplist==[] ? [path]
              :               
               let(
                   fragments = [for(i=idx(duplist)) select(path, duplist[i], select(duplist,i+1))]
               )
               len(fragments)==1 ? fragments
                                 : _assemble_path_fragments(fragments)
            ]
    )
    [for(path=splitpaths) list_unwrap(path)];

function _assemble_partial_paths_recur(edges, paths=[],i=0) =
     i==len(edges) ? paths :
     norm(edges[i][0]-last(edges[i]))<EPSILON ? _assemble_partial_paths_recur(edges,paths,i+1) :
     let(    // Find paths that connects on left side and right side of the edges (if one exists)
         left = [for(j=idx(paths)) if (approx(last(paths[j]),edges[i][0])) j],
         right = [for(j=idx(paths)) if (approx(last(edges[i]),paths[j][0])) j]
     )
     let(
          keep_path = list_remove(paths,[if (len(left)>0) left[0],if (len(right)>0) right[0]]),
          update_path = left==[] && right==[] ? edges[i]
                      : left==[] ? concat(list_head(edges[i]),paths[right[0]])
                      : right==[] ?  concat(paths[left[0]],slice(edges[i],1,-1))
                      : left[0] != right[0] ? concat(paths[left[0]], paths[right[0]])
                      : concat(paths[left[0]], slice(edges[i],1,-2)) // execution flow never seems to get here
     )
     _assemble_partial_paths_recur(edges, concat(keep_path, [update_path]), i+1);
````

@adrianVmariano I am not sure what you mean by "loop removal" in this context. I don't see where in the path assembly code an assumption or requirement that paths are closed except for where it appears to ignore the last closing segment in an assembled path - where I added a comment to the line above that execution never seems to get there.

The only potential problem I can think of is when two contours share one vertex, which could happen if a pixel corner is exactly at the isovalue. You'd have four segments all radiating from a single shared endpoint. There are three ways to connect the segments: two of the ways are acceptable (paths bend away from the shared point), and one is incorrect (paths cross the shared point). All that matters here is avoiding a crossing. Whether the two paths are closed shouldn't matter.

>Does two passes add significantly to the time compared to a one pass approach?

No, it shouldn't. And I am not seeing the need. I'd like to see an example that causes the path assembly to mess up. If you can imagine some set of contour paths that would do this, it's simple enough to make a height map manually to reproduce those contours, as I did in the example above. There, you expressed concern about paths intersecting two different boundaries, so I contrived those contours so the closing segment would cross the path.

**Comment by revarbat** [comment]
So, am I correct that we have three separate implementations of path fragment assembling code that are just slightly different from each other?

**Comment by amatulic** [comment]
@revarbat - yes, the path assembly here was made especially for the contour function.

However, after experimenting with the example Adrian gave in chat, I am convinced that what we really need here is a path assembly function that doesn't make any assumptions about whether paths are open or closed. It should just blindly assemble segments until no more can be joined. After that, any crossing paths (unlikely but possible) should be uncrossed. That is, if we have points a, b, c, d, e arranged in a cross like this:
````
    a
    ↓
b → c → d
    ↓
    e
````
Then paths `[a,c,e], [b,c,d]` are invalid because they cross. The resolution is to make them `[a,c,d], [b,c,e]`.

The segments given to the assembly are such that all paths, whether open or closed, are already completely described by the segments in the collection.

**Comment by amatulic** [comment]
I found something disturbing. The path assembly, even for closed paths, is missing some segments. Consider the following code, which goes through the steps contour() goes through without actually calling contour():
````
include <BOSL2/std.scad>
include <BOSL2/isosurface.scad>

ff = function(x,y) cos(5*atan2(y,x));
pixsize = [3,3];
bbox = [[-7.5,-7,5],[7.5,7.5]];
use_centers=true;
isovalue=0;
closed=true;
smoothpasses=0;

pixels = _contour_pixels(pixsize, bbox, fieldarray=undef, fieldfunc=ff,
    pixcenters=true, isovalue=isovalue, closed=closed);
segtable = use_centers ? _MTriSegmentTable : _MSquareSegmentTable;
seglist = _contour_vertices(pixels, pixsize, isovalue, segtable);
region = _assemble_partial_paths(seglist, closed);
smoothregion = _region_smooth(region, smoothpasses, bbox);
path = smoothregion;

rainbow(path) stroke($item, closed=true,width=.2);

right(20) {
    rainbow(seglist)
        stroke($item, closed=false, width=0.2, endcap2="arrow2");
    color([1,1,1,0.99]) for(x=[bbox[0].x:pixsize.x:bbox[1].x-0.01], y=[bbox[0].y:pixsize.y:bbox[1].y-0.01])
        translate([x+0.1,y+0.1,-0.1])
            cube([2.8,2.8,0.1]);
}
````
This code displays the assembled path on the left, and the segment soup on the right. There's a piece left of center that's missing:
image

**Comment by adrianVmariano** [comment]
Alex, the problem is that dealing with crossings is hard.  You're welcome to devise a path assembly algorithm that can detect them as it goes and solve the problem more efficiently without making any assumptions.  The current implementation separates the two things because detecting crossings is a less local process.  We could maybe detect crossing points in advance by finding all vertices that appear more than twice.   That has a downside of running vertex comparison on twice as many points.   As things stand, there's no way to detect crossings as the path is being assembled.  That's why it's handled as a second step after path assembly is finished.  

Revar, what we have really are 3 implementations, 2 of which are almost the same and one of which is kind of different.   We can just use _assemble_path_from_fragments for this situation.  But it's vastly slower.  The way that you wrote that path assembly you're meant to have a small number of fragments that are broken at intersection points.  What we have in this case is a set of 3 point path pieces.  And paths can never cross in this case, they can only intersect at points, so computing path crossing isn't needed.   So the code I supplied assembles the paths first using a method very much like what's done for vnf_halfspace.  (Those two codes could maybe be combined;  the vnf_halfspace code assumed the input was a list of edges instead of a list of paths.)  After that it finds all duplicate vertices because any duplicate vertex means a path crossing.  It then splits the path on those vertices and passes the result to the fragment assembly to get a loop-free result.  Crossing are assumed to be impossible in vnf_halfspace, since we had a manifold going in.  

Note that a bug in the first phase of path assembly could easily omit an input path segment, though I thought I had this debugged.   I was puzzled for a long time by segments that apparently have no arrow head.  And then I realize they make turns but touch at points.  The shuffle option on rainbow might make this more clear.  This shows that handling crossings at the assembly stage would require breaking the input paths, because the crossings are all in the middle of input path fragments.

**Comment by amatulic** [comment]
There are **no** crossings in the middle of input path fragments. That just isn't possible with the path rules encoded in the lookup table. Two contour paths intersecting the same pixel cannot cross, ever.

An input path can have as many as 4 segments. If two paths intersect a pixel, they would have 2 segments each, I believe.

The existing path assembly function would be fine if used just with closed paths, and then I remove the bounding box edges. I can do that, and it would be fairly tidy. I haven't yet found any place where actual path crossings happen. Most concerning to me, however, is that it's dropping some input paths.

**Comment by adrianVmariano** [comment]
Hmmm.  Yeah, right, those are points of contact, not crossings.   Are crossings possible at all?  That is, can we have more than two paths terminating at the same point?

**Comment by amatulic** [comment]
This is the only way I can imagine it's possible to have a crossing. In the example I mentioned above, with four segments [a,c], [b,c], [c,e], [c,d] (all sharing point c) connected start→end like this. You can imagine that the paths are closed with d connected to a through another path and e connected to b through another path.
````
    a
    ↓
b → c → d
    ↓
    e
````
If, and only if, point c is a corner shared by four pixels, **and** it corresponds to the isovalue, then the path assembly could pick two ways to make two paths out of this. [[b,c,d],[a,c,e]] cross, and [[b,c,e],[a,c,d]] don't.

Even then, I don't know if this situation is possible because each pixel would generate their own paths, likely in opposite directions. I need to contrive a case that actually has this arrangement of corner values and see what happens.

**Comment by adrianVmariano** [comment]
```
ff = function(x,y) norm([x,y])*sin(5*atan2(y,x));
path=contour(ff, bounding_box=16, pixel_size=1,isovalue=0,closed=false);
stroke(path, closed=false,width=.1);
```

The above case has me puzzled.  I've fixed it to it's zero in the center, and the solution really should be a bunch of lines radiating from the center.  Instead the start well away from the center and some get closed off even though I gave closed=false.  The isovalue is exactly achieved at (0,0) which should be on pixel corners.  

image

Some of the centers of the wedges are substantially more than a pixel away from the true corner point, which seems surprising.   Hmmm.  Maybe that's because of smoothing.

Set pixel size to 0.1 and I then get this:  

image

Are the closing segments being added during loop removal?

**Comment by adrianVmariano** [comment]
Can there be a path fragment that is valid but has its start point the same as its end point?  Such path fragments are discarded.

**Comment by adrianVmariano** [comment]
Does "fix the corner values so they never equal the isovalue" mean that you break my example because you are modifying the origin so the parts don't meet there?

**Comment by adrianVmariano** [comment]
With loop removal disabled your prev example does this:

image

**Comment by adrianVmariano** [comment]
I think maybe you changed the case where a fragment is placed between two paths in an effort to prevent paths from being closed and then didn't change it back.  

`left[0] != right[0] ? concat(paths[left[0]],slice(edges[i],1,-2), paths[right[0]])`

With that fix deployed the above becomes:

image

So now there are loops, but no missing segments.  The data explicitly contains loops so getting them isn't a big shock.

**Comment by amatulic** [comment]
> Can there be a path fragment that is valid but has its start point the same as its end point?

Yes. There are two cases where, if the center of the pixel is evaluated as a function instead of a mean of the corners, **and** the center is higher or lower than **all** of the corners, **and** the isovalue is in between the center value and all the corner values, then there is a closed 4-segment path inside the pixel.

**Comment by adrianVmariano** [comment]
If you want to remove **nothing** then you need to delete the 3rd line which removes input paths that have equal endpoints.  

The case where you say "never gets here" should happen exactly once per output closed path, so that's a little odd that it never gets there.  Are you sure?  That's the case where the path fragment closes off one of the paths in the output list.   I tested it on your previous example and there are 2 components and it runs that line twice as expected.  And if you want it to produce a closed output (duplicate endpoints) then that "never gets here" line should have -1 instead of -2 as I think previously noted.  The -2 was to avoid adding the duplicating point.

**Comment by amatulic** [comment]
Remove which third line? `: right==[] ? concat(paths[left[0]],slice(edges[i],1,-1))` ?

The comment about "never seems to get here" was the result of not noticing any difference between using -2 or -1 as the argument. I didn't do extensive testing on that.

**Comment by adrianVmariano** [comment]
```
function _assemble_partial_paths_recur(edges, paths=[],i=0) =
     i==len(edges) ? paths :
     let(    // Find paths that connects on left side and right side of the edges (if one exists)
         left = [for(j=idx(paths)) if (approx(last(paths[j]),edges[i][0])) j],
         right = [for(j=idx(paths)) if (approx(last(edges[i]),paths[j][0])) j]
     )
     let(
          keep_path = list_remove(paths,[if (len(left)>0) left[0],if (len(right)>0) right[0]]),
          update_path = left==[] && right==[] ? edges[i]
                      : left==[] ? concat(list_head(edges[i]),paths[right[0]])
                      : right==[] ?  concat(paths[left[0]],slice(edges[i],1,-1))
                      : left[0] != right[0] ? concat(paths[left[0]],slice(edges[i],1,-2), paths[right[0]])
                      : concat(paths[left[0]], slice(edges[i],1,-1)) // Keep extra point to mark closed paths
     )
     _assemble_partial_paths_recur(edges, concat(keep_path, [update_path]), i+1);
```

The third line meant the third line...from the top.  There's the revised version.  Weird that you didn't notice a difference between -1 and -2 in that line, because I noticed that I got a duplicate endpoint with -1 and not with -2.

amatulic:
I left -1 as the final argument to the final line, leaving the duplicate point for the passthrough function to remove only if closed=true.

PR #1639, opened 2025-04-17
add plot3d

adrianVmariano:


amatulic:
@adrianVmariano I think you need some explanation of zclip. It isn't clear whether it's clipping the z values of the function before zspan, or setting a vertical clip boundary in CAD units for rendering.

Looking at the code, it appears you're clipping the function first, and then scaling it. That isn't intuitive to me.

Throughout the wiki there's inconsistent spelling of words ending in "ward" (backward, forward, upward, etc.). Change "towards" to "toward" here, which is the more common American usage ("towards" is more British but even there they aren't consistent).

amatulic:
Well, I may want clipping for effect, like how some metaball examples make use of bounding box clipping deliberately to create flat areas. Two approaches: I could either (a) scale my function to fit between two values in zspan that are past clipping bounds that are fixed in CAD space, or (b) clip my function first using the z-units of the function so it has the shape I want, and then scale those values to the size I want in CAD space.

Which one happens first? It isn't clear from the docs what these two parameters actually do. Do the clip boundaries get scaled along with the function? Or are they fixed in position regardless of how the function is scaled?

adrianVmariano:
Since clipping and scaling are exclusive itydoes not makes sense to ask which one is first:  you can only do one of them.  

If I were to make both possible I would think clipping needs to be first.  Should I do that?

amatulic:
Oh, I saw the assert() there now preventing both from happening at once.

Hmm. That's OK then. Simpler too and easy to understand if stated up front. I can always build in a z scale and z offset into the function itself if I want to do both at the same time.

However, the docs as written made me expect expect them to work together. It would be worth clarifying in the parameter docs, something like this:
````
//   zclip = A vector `[zmin,zmax]' that constrains the output of function to these bounds. Cannot be used with `zspan`.  
//   zspan = A vector `[zmin,zmax]` that rescales and shifts the function output values to fit exactly within these bounds, so that the the minimum value of `f` appears at `zspan[0]` and the maximum at `zspan[1]`. Cannot be used with `zclip`.
````

**Comment by amatulic** [comment]
So about `bilerp()` which says to see also bilerpn....

It seems like there's a hole, needing a `bilerp_mxn()` to return m&times;n values interpolated between the four points.

**Comment by adrianVmariano** [comment]
Should I actually add bilerpnm or whatever?

**Comment by amatulic** [comment]
I don't know how useful bilerpmn would be but it seems like an obvious thing to have given that bilerpn exists.

On the other hand, it isn't hard for someone to implement. It's just two nested loops of lerpn.

**Comment by amatulic** [comment]
I noticed another thing. If `echo_viewport()` is going to be public-facing, it might be worth noting that your viewport needs to have a 4&times;3 aspect ratio for the output to work correctly in the wiki. This has bitten me in the past.

**Comment by adrianVmariano** [comment]
Yeah, I size the window to match an example from the docs.  Is there a simple way to force the viewport to a desired aspect?   I got tired of constantly writing a line of code to print these things out and figured it may as well be in the library.

**Comment by amatulic** [comment]
I don't know of any way to force the viewport to a specific size. However, it's helpful to see the viewport dimensions in the lower left frame of the OpenSCAD window. I just size it to be 600x450 on my low-res external monitor, or 800x600 on my hi-res laptop monitor. It's hard to get it exact but being a few pixels off is close enough. Where things get messed up is if I resize the whole OpenSCAD window, then the viewport changes size too, and I forget to correct the size.

**Comment by amatulic** [comment]
If those recent checks don't pass, here are some suggestions for the next commit:
* New text in rounding.scad:
  * "because it internally uses {{offset()}} which will expand those sharp corners into very long single segments that produce incorrect result." - 
  * Make that "an incorrect result" or "incorrect results". I would also change "will expand" to "expands" and remove the meaningless intensifier "very" per [Mark Twain's sage advice](https://www.goodreads.com/quotes/2913-substitute-damn-every-time-you-re-inclined-to-write-very-your), but that's your call.
* In shapes3d.scad, you might want to put "(deprecated; use {{plot3d()}})" in the synopsis section of heightfield() so that people can see it's deprecated without having to discover this by looking at a warning message after using it.
* Should we enclose deprecation messages between `*****` elsewhere? If so, you may want to search for the string "deprecated" in vnf.scad and beziers.scad (in which I just introduced deprecation warnings) and change them, and include them in this PR too.

**Comment by adrianVmariano** [comment]
Note that heightfield is no longer documented.  If you look carefully there are three '///' in front of its doc text.  In fact, all the recent check failures have been because of references elsewhere in the docs to heightfield().  

I followed the form of the deprecate() module with the asterisks.  I don't think it's that important.  Those messages are all going to vanish when we release the stable version.

**Comment by adrianVmariano** [comment]
The revised text still seems funny to me.  It switches from "internally uses" to "would" which doesn't seem right.  Like it's less counterfactual than that?  Or some transition is needed?  "If you pass sharp corners it would make long segments" doesn't seem right.  I guess maybe "If you were to pass sharp corners it would make long segments" is the fix.   But I don't like that.  What happens if you *do* pass sharp corners.  To me it seems like "Don't pass sharp corners.  If you do pass sharp corners it will make long segments." is correct.   The "will" is not acting as a future here but something else.  

That error check on z is broken.  The other thing is that angle is a range/list since it defines the evaluation points for the function.   You're actually supposed to get an error if you only give a single value.   And your function's first parameter (the theta parameter) isn't properly scaled for a full circle of rotation.  It's supposed to go from -10 to 10.  This actually is the case where the arclength parameter seems attractive, but it will distort the shape on a sphere (or cone).   There is a bug with how thresholding for small r is done that thresholds the function at zero instead of the final r value.  Also `style` isn't passed through.  So fixing those things:

```
a=5;b=10;
g = function(x,y) sin(360*(x/18)/a + 200 * sin(360*y/b));
plot_revolution(g, [0:2:360], path=arc(100, r=10, start=-89, angle=178),style="min_area");
```

image

Seems like triangulation could be better.   What do you think?  The stock surface normal function adds 0.7s to the run time, which is a significant fraction, but not exactly intolerable.  I don't know if there's a faster way to compute the normals.   But even if it's slow if it works better, it's probably worth having it.  Users could turn it on only for final renders, for example.   

I thought I had captured the examples I posted to the chat.  Is there anything I missed that didn't make it in as an example already?  

The question about determining angle form the texture is this:  You're placing a texture onto a wedge portion of a cylinder.  You want the texture to appear at its true aspect ratio, e.g. because it's a bitmap image.  So what angle do you pass to rotate_sweep to achieve that outcome?  The answer is you have to do something like `rotate_sweep([[r,-height/2],[r,height/2]], texture=pic, angle =picwidth/picheight*360*height/2*PI*r)` where the calcluation of angle produces 1:1 pixels on the cylinder mapping from the original image.

adrianVmariano:
image

amatulic:
How about simply "passing sharp corners would make long segments..."

The wiggly sinewave sphere turned out lovely. The effect of horiz=true is evident only at the poles.

PR #1664, opened 2025-05-01
clipped circles for offset_sweep

adrianVmariano:


amatulic:
I have to say I'm surprised and pleased. Pleased that this is the kind of rounding I always wanted, and surprised that the change to the "machinery" involved just two lines of code according to the diff, with the rest of the changes being documentation and input handling.

PR #1667, opened 2025-05-04
Improvements to line_from_points() and plane_from_points()

amatulic:
Changes to geometry.scad to make API more consistent:
* Functions that used a `fast` parameter to skip collinearity or coplanarity checks now use `check_collinear` or `check_coplanar` as needed. This affected `line_from_points()`, `plane_from_points()`, and `plane_from_polygon()`.
  * Note the logic is opposite; `fast=true` corresponds to `check_coplanar=false`.
  * ~Should I add deprecation code for `fast`?~ **- Done.**
* `line_from_points()` now returns an actual best-fit line instead of a line connecting two arbitrary points, if the collinearity checking permits.
* In both `plane_from_points()` and `line_from_points()`, the default behavior is now to return the best-fit plane or line, skipping coplanar/collinear checking. This is the opposite of what it was before.
* `plane_from_polygon()` still checks for coplanarity as it did before.
* Added example for `line_from_points()`, improved example for `plane_from_points()`.
* Added a newline in front of all assert messages, and minor grammar tweaks.

Changes to coords.scad:
* ~Commented out asserts assocated with calls to `plane_from_points()` in `project_plane()` and `lift_plane()` because by default the coplanarity checking no longer occurs, and it is not required to have coplanar points to define a plane for projection or lifting.~ **Reverted to original behavior, added `check_coplanar` to these calls instead** per discussion below.
* Added a newline in front of all assert messages, and minor grammar tweaks.

### DIFF
--- a/coords.scad
+++ b/coords.scad
@@ -39,9 +39,9 @@ function point2d(p, fill=0) = assert(is_list(p)) [for (i=[0:1]) (p[i]==undef)? f
 // Arguments:
 //   points = A list of 2D or 3D points/vectors.
 function path2d(points) =
-    assert(is_path(points,dim=undef,fast=true),"Input to path2d is not a path")
+    assert(is_path(points,dim=undef,fast=true),"\nInput to path2d is not a path.")
     let (result = points * concat(ident(2), repeat([0,0], len(points[0])-2)))
-    assert(is_def(result), "Invalid input to path2d")
+    assert(is_def(result), "\nInvalid input to path2d.")
     result;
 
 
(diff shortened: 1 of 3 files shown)


adrianVmariano:
I'm not sure what I think about having the functions work by default on non-coplanar or non-collinear points.  I feel like the main use is for points that are supposed to be coplanar and having it work anyway on arbitrary data could be a source of confusing bugs.   My intended use case is where the situation where lift_plane is the inverse of project_plane, meaning that all the inputs are actually located on the plane.    Is there a way to enforce coplanarity in lift_plane and project_plane?  

Probably a good idea to add deprecation message for fast.

amatulic:
> Is there a way to enforce coplanarity in lift_plane and project_plane?

Yes, you just add `check_coplanar=true` where they call plane_from_points(). However, I disagree that project_plane should enforce this. I was planning to use project_plane() specifically to project onto a best-fit plane.

adrianVmariano:
I meant is there a way for a user to get the historical behavior that project_plane fails if the input is not coplanar.

amatulic:
I've added the `check_coplanar=true` parameter to where project_plane() and lift_plane() call plane_from_points(), and reverted the associated documentation back to remove saying that the points needn't be coplanar.

I'm not sure what you mean by "historical behavior." If you mean the history of a fail, you have the trace information in the console. If you mean the development history of project_plane() and lift_plane(), they are now back to working like they were originally.

adrianVmariano:
You could have check_coplanar as an option on project_plane and lift_plane.   I'm not sure what the typical use cases are for these, really.  You can always compute the plane itself and pass that, not the 3 points.   One could even argue that as an API choice that might make more sense.  Get your plane how you like.  But it seems late to change that.

amatulic:
> You can always compute the plane itself and pass that, not the 3 points.

So, if a vector of four values is passed in, project_plane() and lift_plane() would just use that and not need to call plane_from_points().

That should be an easy change, and it makes a lot of sense. I could call plane_from_points() myself so that I could save the plane coefficients for later use, and simply pass them into project_plane().

adrianVmariano:
So is this ready for merge?

amatulic:
Yes, should be ready now.

PR #1669, opened 2025-05-04
Make offset() return [] if offset path is degenerate and error=false

amatulic:
Changes to regions.scad:
* Added `error` parameter to `offset()`. If the offset path is degenerate and `error=false`, an empty path `[]` is returned. Default is `error=true`, which causes an assert as usual.
* Edited some examples to cause their code to appear on the right side of the figure.
* Fixed a couple of typos, minor grammar and wordsmithing.
* Added newline preceding each assert message.

In the diff, the actual code change to handle the `error` parameter starts at line 1029. Everything else in there are inconsequential changes.

I've tested this and it works like I expect.

What I don't quite understand is how the quality setting works. For example, this code to produce an hourglass shape returns `[]` for an inner offset **only** when `quality=2`, and not for any other value like 1 or 3.
````
include <BOSL2/std.scad>
path = [[-10,-10], [-10,10], [0,3], [10,10], [10,-10], [0,-3]];
inpath = offset(path,delta=-5, same_length=false, quality=2, error=false, closed=true);
stroke(path, closed=true, color="green");
if(len(inpath)>0) stroke(inpath, closed=true, color="red");
````

### DIFF
--- a/regions.scad
+++ b/regions.scad
@@ -865,13 +865,13 @@ function _point_dist(path,pathseg_unit,pathseg_len,pt) =
 //   r = offset radius.  Distance to offset, rounds over corners.
 //   delta = Distance to offset with pointed corners.
 //   chamfer = Chamfer corners when you specify `delta`.  Default: false
-//   closed = If true, path is treated as a polygon. Default: True.
-//   check_valid = Perform segment validity check.  Default: True.
-//   quality = Validity check quality parameter, a small integer.  Default: 1.
-//   error = If true, assert an error if offset path is degenerate. If false, return an empty list `[]` for a degenerate path. Default: true
+//   closed = If true, path is treated as a polygon. Default: true
+//   check_valid = Perform segment validity check.  Default: true
+//   quality = Validity check quality parameter, a small integer.  Default: 1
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by adrianVmariano** [comment]
```
degenerate && error ? assert(false, "\nOffset of path is degenerate.")
    : degenerate && !error ? [] // return empty path
```

I would write this as

```
assert(!(degenerate && error), "\nOffset of path is degenerate.")
degenerate ? []
```

but maybe you thought the above was more clear?

amatulic:
No, your logic is better. I'll submit another change, and I noticed a couple more typographical corrections I can make (capitalizing "true" inconsistently).

amatulic:
In that code snippet above, setting quality to 4 or 5 give some strange results. Quality of 2 or 6 give me what I expect.

adrianVmariano:
Yes, the test in offset() is heuristic and definitely not robust.   In the case at hand, for you to get the result you want, I think the tests have to fail (by which I mean, produce incorrect results), because the offset path is not actually degenerate yet.  It's a region with two triangles.   But that case can't be handled.   So you can't expect a good answer really no matter what.  It's either going to tell you it's degenerate, which is wrong, or give you a path that connects the degenerate components, which is wrong.

amatulic:
Ready to merge.

PR #1670, opened 2025-05-05
add bot_corner to teardrop2d()

adrianVmariano:


amatulic:
* "The segments of the round section of the teardrop" - verb is missing, append "are" to that.
* Shouldn't the Usage section be above the description?
* "When called as a function" should be moved to the beginning of the description before or after "When called as a module"

I'll leave it up to you to decide what to do with the mixed present and future tense, and "note that".

PR #1691, opened 2025-05-21
Improvements to img2scad.html, fix contour() topic

amatulic:
In img2scad.html:
* Added size value to top of .scad output file
* Added sharpening filter
* Tweaked UI
* Fixed artifact that appeared on image edge after applying Sobel edge detection filter, by making the convolutions use edge reflection

In isosurface.scad:
* One comment line corrected. The `contour()` module/function was classified under the wrong topic.


amatulic:
The method for sharpening an image is to subtract the blurred image from the original. Therefore, blur before sharpen. Blurring gives you the low-frequency components of the image, and if you subtract that from the original, you are left with high-frequency (sharper) components.

If you already blurred the image, then the sharpening uses that instead of blurring it internally. If you didn't blur the image yourself, it blurs the image internally before computing the difference.

In reality, the three filters would likely be used independently, but if you set more than one of them, there has to be a cascade. For both sharpening and edge detection, blurring is the necessary first step.

Edge detection can work with the blurred or the sharpened one, because either way the blurring operation has already been done. Sharpening enhances high-frequency components, and edge detection would enhance them further.

amatulic:
Please upload an example file to test it on. Also do you think it's best if the three filters were all independent and didn't interact? If so, I'd just put radio buttons in front of each one so you can select the one you want.

RAMilewski:
Netfool

Note what happens to the background if you sharpen it.
Yes, I think the radio buttons would be a good idea.

amatulic:
I see. That image is already sharp.

You can see what's happening if you increase sharpening all the way to 20. You still have a white border next to the black outline.

What this filter does is enhance the gradient (change in brightness) and then subtract low-frequency components. At the outer boundary of the figure in the drawing, you have a black edge next to a white background. This change is amplified, so then you have a "blacker" edge surrounded by a small "whiter" region. Internally, the brightness is just a number, and "blacker" is less than 0 and "whiter" is greater than 255, which cannot be displayed. After the sharpening operation is done, the new brightness range is rescaled back to the range 0-255. This results in the regions near the edge remaining white, but because they were greater than the white background before rescaling, the background is made grayer.

Notice that the same thing happens with the black stripes on the headpiece. They turn from black to dark gray, because the black outline adjacent to the white background was made "blacker" (negative) internally before the brightness rescaling.

amatulic:
Version 12 of img2scad.html committed.

I'll add that the gray background you see from sharpening would affect only the height of the background in the texture relative to everything else. The edge of the figure would have a slight ridge before dropping in elevation to the black level.

Issue #1679, opened 2025-05-11
$slop used incorrectly for threading and screws, or $slop doesn't work at small scales


coryrc:
**Describe the bug**

Following the [$slop calibration](https://github.com/BelfrySCAD/BOSL2/wiki/constants.scad#constant-slop) does not result in properly-sized holes for `#6-32` holes or `M7` nuts.

I have a Bambu A1 Mini, using PETG, and am printing some self-tap and clearance holes for `#6-32` screws. I have the following diameters:

|$slop | self-tap dia | self-tap measured | clearance dia | clearance measured |
| ----- | ------------ |  ------------------- |  --------------  | ---------------------- |
|0.00 | 2.94 | unk | 3.91 | unk |
|0.22 | 3.38 | 3.67 | 4.35 | 4.45 |

In both cases, my actual holes are *larger* than specified, which is inconsistent with the intended use of `$slop`.

Yesterday I printed the slop tester and calculated 0.22 for slop. Using the raw threading.scad and with a printed M7 buttress thread and `$slop` = 0.22, I created working nut and screws printed vertically. Some numbers from that print:

| $slop | nut innermost diameter | nut innermost measured |
| ------ | -------------------------- | --------------------------- |
| 0.00 | 5.53 | unk |
| 0.22 | 6.38 | 6.22 |

However this just happened to work, because the `tol_gap` for a normal M7 in the `screws.scad` library would be about 0.8, which is passed to `screw()` as `shaft_undersize=-0.8` so a normal fit would be 6.33mm innermost diameter and setting `$slop = 0.22` *coincidentally* got a working value. (The `threading.scad` library does not have any tolerance adjustment besides `$slop`). Had I correctly compensated diameter for fit, then the nut would have been far too large.

In this case, my actual holes are *smaller* than specified, which is consistent with `$slop`, just way more than it should have.

**Code To Reproduce Bug**
```
module clamp_block() {
    diff() cube([1.9,6.5,12],anchor=BOT+RIGHT) {
        position(TOP+RIGHT) cube([2,6.3,1.9], anchor=TOP+LEFT);
        tag("remove") up(tapping_hole_z) position(BOTTOM) screw_hole(screw_info("#6-32"), tolerance="tight", l=10, orient=LEFT, $slop=0.22);
    }
}
module tapping_hole() {
    screw_hole(screw_info("#6-32"), tolerance="self tap", l=10, orient=LEFT, $slop=get_slop()/2);
}

[...]

PR #1699, opened 2025-05-30
Change $slop docs and tester

coryrc:
Incorporating findings from Issue #1679

Describing why based on the diff on my screen, starting top to bottom:

Removing XY vs Z axis, because the printing direction isn't known for the parts. Even precise CNC machining still has tolerances and I tried to write a simple way of passing that along.

Adding the limitations we talked about in issue #1679.

Added some more written examples of where slop should be added. Now that I'm writing this, these would be better as an example.

I can't add a new line to the first "Calibration" line because that'll make a new number. I thought this was good info to add though.

Moved caveat to Limitations section.
Added line about filleting/chamfers now that I made the tester cut off corners, so you can test both ways.

Also fix a typo later in doc

slop_tester_square

### DIFF
--- a/constants.scad
+++ b/constants.scad
@@ -19,15 +19,36 @@ _UNDEF="LRG+HX7dy89RyHvDlAKvb9Y04OTuaikpx205CTh8BSI";
 // Synopsis: The slop amount to make printed items fit closely. `0.0` by default.
 // Topics: Constants
 // Description:
-//   A number of printers, particularly FDM/FFF printers, tend to be a bit sloppy in their printing.
-//   This has made it so that some parts won't fit together without adding a bit of extra slop space.
-//   That is what the `$slop` value is for.  The value for this will vary from printer to printer.
-//   By default, we use a value of 0.00 so that parts should fit exactly for resin and other precision
-//   printers.  This value is measured in millimeters.  When making your own parts, you should add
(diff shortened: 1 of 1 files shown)


revarbat:
I like the corner cutouts, but shouldn't they be on all four corners?

coryrc:
This lets you rotate 180 degrees to check if you have bulging corners (aka uncalibrated pressure advance) affecting slop. (it's written in the docs).

constants.scad
```diff
@@ -44,25 +65,36 @@ _UNDEF="LRG+HX7dy89RyHvDlAKvb9Y04OTuaikpx205CTh8BSI";
 //   The correct hole should hold the plug when the long block is turned upside-down.
 //   The number in front of that hole will indicate the `$slop` value that is ideal for your printer.
 //   Remember to set that slop value in your scripts after you include the BOSL2 library:  ie: `$slop = 0.15;`
-//   .
-//   Note that the `$slop` value may be different using different materials even on the same printer.  
+//   If trying to test the proper `$slop` value for a mating cube and matching socket without any filleting or chamfers, you can turn the plug 180 degrees and follow the above procedure again.
```
coryrc, line 68: This is where I mention slop for square fit.

PR #1700, opened 2025-05-30
Change threading and screws to use 2*$slop

coryrc:
Per [my argument](https://github.com/BelfrySCAD/BOSL2/issues/1679#issuecomment-2923309921), I believe this is the proper method. There is only one mating surface for each radius, so by following the documentation for `$slop`, we should only be adding one `$slop` for radius and not two.

This is backward-incompatible.

### DIFF
--- a/screws.scad
+++ b/screws.scad
@@ -593,7 +593,7 @@ module screw(spec, head, drive, thread, drive_size,
             assert(is_undef(user_thread_len) || (is_finite(user_thread_len) && user_thread_len>=0), "Must specify nonnegative thread length");
    sides = max(pitch==0 ? 3 : 12, segs(nominal_diam/2));
    rad_scale = _internal? (1/cos(180/sides)) : 1;
-   islop = _internal ? 4*get_slop() : 0;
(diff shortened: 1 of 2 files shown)


revarbat:
One slop per radius is for **TIGHT** fits, which fails to let a screw screw/unscrew at larger radiuses, and I imagine will fail to let you insert the screw at all, and/or break the screw at low radiuses.

coryrc:
> One slop per radius is for **TIGHT** fits, which fails to let a screw screw/unscrew at larger radiuses, and I imagine will fail to let you insert the screw at all, and/or break the screw at low radiuses.

That's why screws.scad has a [tolerance system](https://github.com/BelfrySCAD/BOSL2/blob/d0bdc25da92dbeb1ddabb464429666926e66cf28/screws.scad#L129) using ISO standards for making working screw threads.

adrianVmariano:
The slop system is imperfect and maybe it's even wrong, but we don't want to change it because of subtle backwards compatibility issues that such changes would introduce.

PR #1711, opened 2025-06-12
Add keep= argument to partition

coryrc:
If you only want one half of the partition, this argument makes it easy (and efficient).

The following does not work as an alternative:

```
partition(..) if($idx==0) the_child();
```

because the intersection{} then only has one object and returns the mask instead of nothing.


adrianVmariano:
I think you need to disable lazy union in your dev snapshot or use the stable version. The lazy union feature breaks a variety of things and the word is it will not be in the next stable release.  With lazy union off, you can just do `if ($idx)` and this extra feature isn't needed.

adrianVmariano:
Since this can be done using `if ($idx==0)` I'm seeing it as unnecessary, so unless you have a further argument supporting the need for this---did I miss something?---I think we are going to close it.

coryrc:
I've tried making a simple repro to file with openscad and I have not been able to.

I'm updating this with a warning about lazy union and I came up with a simpler workaround.

coryrc:
Actually, I think the existing code is relying on undefined behaviour. `intersection()` is documented as "Only the area which is common or shared by all children is retained.". If there are no children (the else clause) then `children()` should be an empty set and `intersection()` should only have one child: the mask. So I still think this is a good change to make, but I'm going to make a different PR for a doc update alternative.

adrianVmariano:
The question is whether no children is an empty set or whether no children is just no object.  There is no such thing as "undefined behavior" in openscad.  Defined by what?  The program defines its behavior.  The documentation is basically a description of the behavior.  (It's a user editable wiki, so you can update it to reflect your observations.)   It's not like there's a spec that it's supposed to adhere to.   The lazy union feature under some circumstances changes this behavior about whether "no object" becomes an empty set or just no object.

coryrc:
The docs+union recommendation is overall better, so withdrawing this.

On the discussion topic, nothing defines how `intersection() { children(); cube(); }` should work when there aren't children, so it's undefined behaviour. From the AST there are two children (so $children would be 2 inside intersection() were it written in scad) so perhaps intersection() should be checking every AST child has a geometry (or whatever word).

Anyway, with that definition of UB there's quite a bit of undefined behaviour :-)

adrianVmariano:
@coryrc I have evidently failed to clearly make my point.  The only DEFINITION of openscad's behavior is the OpenSCAD program.  Whatever it does is the definition of the behavior.  It does something when you do intersection () { children(); cube();} when there aren't children, ergo that is the definition of the behavior.  

The notion of "undefined behavior" is impossible when the only manual is a publicly editable DESCRIPTIVE manual written by the users.   I rewrote the section on the search() function a few years ago, for example, because it was inconsistent with the actual program behavior.  I tested a whole bunch of cases in an effort to correctly document what actually happens.  If we had a definitional manual such an inconsistency would mean that I'd found bugs in the search() function.  

I could go edit the manual right now to add details about the case in question.  Would it then suddenly become "defined behavior?"

PR #1712, opened 2025-06-12
Fix typo in skin.scad

coryrc:
Introduced in https://github.com/BelfrySCAD/BOSL2/commit/0b6931bce78aaf439e440a80917cc51646559b77

There is no `path` argument and never was.


adrianVmariano:
I've fixed this in a separate PR, which seems easier than getting the tests to run here.

PR #1721, opened 2025-06-17
Doc how to only keep half a partition

coryrc:
And workaround for Lazy Union change in behaviour with `intersection() { children(); mask(); }` when there are no children.

Alternative to #1711

### DIFF
--- a/partitions.scad
+++ b/partitions.scad
@@ -541,6 +541,9 @@ module partition_cut_mask(l=100, h=100, cutsize=10, cutpath="jigsaw", gap=0, cut
 //   partition(size, [spread], [cutsize], [cutpath], [gap], [spin], [$slop=]) CHILDREN;
 // Description:
 //   Partitions an object into two parts, spread apart a small distance, with matched joining edges.
+//   If only one side of the parition is desired, $idx can be used on the children i.e.
+//   `partition() if($idx==0) cube()` will only give the back partition. This breaks with Lazy Union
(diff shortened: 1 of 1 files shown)


adrianVmariano:
Documenting how to do this is good.

But I'm reluctant to give specific instructions for dealing with lazy union, since it's an experimental development feature that is going to go away.  It's a thing we'd have to chase down and remove from the docs once the next stable release appears.  Note also that a simpler workaround is to put in an explicit union() around your child, as in:

`partition() union() {  if ($idx==0) thing();}`

coryrc:
Are you sure it's going away? I couldn't find any discussion on the github issue or web searching that indicated so.

The union is much better. I will update that. I could also make it the only recommendation and not say the union is needed for lazy union (per outcome of above) thus if someone else has the same problem as me, go to docs, they'll at least see a difference, even if they don't know why.

adrianVmariano:
It is impossible to be sure of anything.  But Marius wrote that lazy union was going away on the OpenSCAD mailing list.  In one minute I couldn't find the message to quote it.   He said that lazy union had been introduced to address performance issues and that manifold addressed all those same issues and made lazy union unnecessary, and that furthermore, lazy union created compatibility issues.   So therefore he said it would be going away.  

I would prefer not to suggest unnecessary stuff (e.g. to unconditionally suggest the use of the union(), which then raises questions about why is that there?)   I guess if you say "if you are using a development snapshot with "lazy union" enabled then...."    I would guess this is not the only place where lazy union breaks something in the library.

adrianVmariano:
Redid this in a different PR to show it in an example and to write more in the active voice.

PR #1743, opened 2025-07-02
shapes3d.scad: create u_shape() module

twitchyliquid64:
I use this module all the time so thought it might be a useful addition.

Lmk if I'm doing this right!

### DIFF
--- a/shapes3d.scad
+++ b/shapes3d.scad
@@ -1890,6 +1890,53 @@ function octahedron(size=1, anchor=CENTER, spin=0, orient=UP) =
     ) reorient(anchor,spin,orient, vnf=vnf, extent=true, p=vnf);
 
 
+// Module: u_shape()
+// Synopsis: Creates the shape of a 'U' or rotated-C character, such as might be used in a C-clamp.
+// SynTags: Geom
+// Topics: Shapes (3D), Attachable
+// See Also: prismoid()
+// Usage: As Module
+//   u_shape(width, height, ...) [ATTACHMENTS];
+// Description:
+//   When called as a module, creates the shape of a 'U' with the specified gap width and height.
+// Arguments:
+//   width = The space between either side of the U.
+//   height = The vertical space from top of the bottom bar of the U, to the top of the part.
+//   thickness = The thickness of the U shape in all dimensions not constrained by width or height.  Default: `8`
+//   ---
+//   anchor = Translate so anchor point is at origin (0,0,0).  See [anchor](attachments.scad#subsection-anchor).  Default: `CENTER`
+//   spin = Rotate this many degrees around the Z axis after anchor.  See [spin](attachments.scad#subsection-spin).  Default: `0`
+//   orient = Vector to rotate top towards, after spin.  See [orient](attachments.scad#subsection-orient).  Default: `UP`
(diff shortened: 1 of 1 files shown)


twitchyliquid64:
```scad
   diff()
   u_shape(20, 20) {
     attach("bore_left", TOP, inside=true, shiftout=1) cyl(r=2, h=10);
     attach("bore_right", TOP, inside=true, shiftout=1) cyl(r=2, h=10);
   };
```

image

amatulic:
I don't know where I'd use this in my work, but I have some observations.
* A simple shape like this is easy enough for anyone to model. Its usefulness, I think, comes from the demonstration of anchoring. As such, it could be an example in a tutorial.
* I am not sure if 3dshapes.scad is the right place, if it's part of a general category of utility parts like clamps.
* The default orientation is sub-optimal for 3D printing a U-shaped object that would undergo flex loads, as with a clamp. Layer bonds are the weakest feature of any print. To withstand flex loads, you would want this shape printed flat on the bed, so that there are no flex loads in a direction that would delaminate any layers.
* Similarly, load-bearing shapes with sharp corners tend to be weaker than curved shapes. It isn't as much of a problem in this case if the two arms of the U are being compressed together except that the corner geometry would create an uneven load distribution across the part. The load on a C-clamp, however, is equivalent to *forcing apart* the two arms of the U, in which case the load becomes concentrated in the sharp concave corners, causing a crack to appear there earlier compared to a rounded shape.

This might be a good tutorial example. I confess I don't understand attachments in BOSL2, and what I see when I test the example here makes no sense to me with respect to where the holes end up.

@RAMilewski [The attach documentation](https://github.com/BelfrySCAD/BOSL2/wiki/attachments.scad#module-attach) also mentions nothing about the parent anchors "bore_left" and "bore_right". They clearly work, but I can't find any documentation about what they do. I looked for these words in some of the tutorial pages too.

twitchyliquid64:
Let me know what you want me to do with it - happy to move to tutorial, a utility parts file etc.

> The default orientation is sub-optimal for 3D printing a U-shaped object that would undergo flex loads, as with a clamp.

The default orientation is one that is consistent with a U-shape. Is the norm for BOSL to assume everything is 3d-printed and to orient all parts that way? Happy to, just wasnt aware if so.

> Similarly, load-bearing shapes with sharp corners tend to be weaker than curved shapes. It isn't as much of a problem in this case if the two arms of the U are being compressed together except that the corner geometry would create an uneven load distribution across the part. The load on a C-clamp, however, is equivalent to forcing apart the two arms of the U, in which case the load becomes concentrated in the sharp concave corners, causing a crack to appear there earlier compared to a rounded shape.

Love this feedback! It sounds like I should be adding a small fillet to the inside corners by default?

> also mentions nothing about the parent anchors "bore_left" and "bore_right"

These are custom anchors for this part. Those anchors represent the 'center' of the U void, rather than the center of the whole part.

adrianVmariano:
I lean towards thinking that this shape is too simple.  Building a robust API that passes all the rounding options of cuboid would make the API kind of complicated, and it's easy enough to make the shape directly in various ways, so it doesn't seem to add enough value to justify its existence as a separate module.

twitchyliquid64:
Very fair, will close this one out!

PR #1747, opened 2025-07-05
Add insertion_mask_slide to dovetail() module

coryrc:
insertion_mask_slide creates a space for the male dovetail to enter tangential to the sliding direction of the joint. It's part of dovetail() because computing the actual cross-section needed involves $slop and slope/angle and taper/back_width and width and height... so it's a lot simpler to add it in to the existing module.

Example and documentation added.

### DIFF
--- a/joiners.scad
+++ b/joiners.scad
@@ -687,7 +687,7 @@ module dovetail(gender, width, height, slide, h, w, angle, slope, thickness, tap
 
     type = is_def(chamfer) && chamfer>0 ? "chamfer" : "circle";
 
-    smallend_half = round_corners(
+    bigend_half = round_corners(
         move(
             [0,-slide/2-extra,0],
             p=[
@@ -700,13 +700,13 @@ module dovetail(gender, width, height, slide, h, w, angle, slope, thickness, tap
         method=type, cut = fullsize, closed=false
     );
 
-    smallend_points = concat(select(smallend_half, 1, -2), [down(extra,p=select(smallend_half, -2))]);
+    bigend_points = concat(select(bigend_half, 1, -2), [down(extra,p=select(bigend_half, -2))]);
     offset = is_def(taper) ? -slide * tan(taper)
            : is_def(back_width) ? (back_width-width) / 2
            : 0;
-    bigend_points = move([offset+2*extra_offset,slide+2*extra,0], p=smallend_points);
+    smallend_points = move([offset+2*extra_offset,slide+2*extra,0], p=bigend_points);
 
-    bigenough = all_nonnegative(column(smallend_half,0)) && all_nonnegative(column(bigend_points,0));
+    bigenough = all_nonnegative(column(bigend_half,0)) && all_nonnegative(column(smallend_points,0));
 
     assert(bigenough, "Width (or back_width) of dovetail is not large enough for its geometry (angle and taper");
 
@@ -715,7 +715,7 @@ module dovetail(gender, width, height, slide, h, w, angle, slope, thickness, tap
 
     // This code computes the true normal from which the exact width factor can be obtained
     // as the x component.  Comparing to wfactor above shows that they agree.
-    //   pts = [smallend_points[0], smallend_points[1], bigend_points[1],bigend_points[0]];
+    //   pts = [bigend_points[0], bigend_points[1], smallend_points[1],smallend_points[0]];
     //   n = -polygon_normal(pts);
     //   echo(n=n);
     //   echo(invwfactor = 1/wfactor, error = n.x-1/wfactor);
@@ -726,8 +726,8 @@ module dovetail(gender, width, height, slide, h, w, angle, slope, thickness, tap
 
             skin(
                 [
-                    reverse(concat(smallend_points, xflip(p=reverse(smallend_points)))),
-                    reverse(concat(bigend_points, xflip(p=reverse(bigend_points))))
+                    reverse(concat(bigend_points, xflip(p=reverse(bigend_points)))),
+                    reverse(concat(smallend_points, xflip(p=reverse(smallend_points))))
                 ],
                 slices=0, convexity=4
             );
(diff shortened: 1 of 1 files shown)


coryrc:
Oh, and I'm not very satisfied with my naming choices and phrasing. Happy to take suggestions.

adrianVmariano:
I don't understand what this does, and the docs are unclear and give no examples.

coryrc:
It does include an example?

But it does need the picture to understand, I think. Attaching now.

```
// Example: Blind housed sliding dovetail (similar to okuri ari).  If the left piece is rotated 90 degrees clockwise, it can be inserted downward with the male dovetail entering the insertion opening, then slid backward, making a 40x40x20 cube with no visible connectors. In this case, the opening is made slightly longer than the male dovetail (17 vs 15) so it's easier to assemble; with tapered dovetails, it's sometimes possible to make it shorter than the length (slide) of the male dovetail.
//   left(30) cuboid([10,40,40]) attach(RIGHT,BOT,align=BACK,spin=90,inset=5) dovetail("male", slide=15, width=20, height=8, slope=2);
//   right(30) diff() cuboid([40,40,10]) attach(TOP,BOT,align=BACK,inside=true,inset=5) tag("remove") dovetail("female", slide=15, width=20, height=8, slope=2, insertion_mask_slide=17);
```

example

coryrc:
I suppose I didn't explain the terminology; AI says:

A "fully housed blind sliding dovetail" is a hidden joint for connecting two pieces of wood. Imagine a wedge-shaped slot (the "dovetail") cut into one board, and a matching wedge-shaped protrusion on another. "Sliding" means these two pieces slide together. "Blind" means you can't see the joint from the outside, and "fully housed" means the joint is completely enclosed within the wood for a clean look.

I could add that to example, but I thought it was already getting wordy and "picture worth 1000 words".

adrianVmariano:
First, I have to apologize: I'm not sure how I missed that you had an example and short documentation text in there.  

Secondly, I'm a woodworker and the joint you've created here is not a woodworking joint.  The terminology around dovetails will tend to describe woodworking joints.   I don't think a "fully housed blind sliding dovetail" exists---that is, AI made it up.  Note that due to the entry hole, it wouldn't actually be blind because you'll see that hole.   A blind (non-sliding) dovetail is a corner joint where the dovetail is hidden behind a miter and no sign of the joint is visible once it is assembled.   Generally in woodworking, a sliding dovetail is a long joint (e.g. a foot long), it's tapered so you can actually assemble it, and if you want to hide it you can only hide it from one side (e.g. the front) because it has to enter from the other side.   If the dovetail is short, there's no reason to add the complication of making it sliding because you can assemble it without that, and if the joint is in the center of a board like in your example, you don't use a dovetail.  

I think we need a better name for the parameter, which we've been talking about in the dev chat.  And the documentation text could probably be expanded.   Maybe `entry_hole_length` but maybe something better can be found.

coryrc:
> Note that due to the entry hole, it wouldn't actually be blind because you'll see that hole.

That's not true -- if you position the example's parts together, you cannot see any part of the joint.

> I don't think a "fully housed blind sliding dovetail" exists---that is, AI made it up.

I made it up, not AI, because just "housed sliding dovetail" was finding tapered shelf dado you slide from the back and hide from the front, but can see from the back (so lacking the "fully blind" part) and some other examples use "blind" to just mean "in normal use". [Dorian Bracht calls it "blind housed dovetail"](https://www.youtube.com/watch?v=wlz9CX0BW4I). This name's not important though.

I just found [Fine Woodworking is using "housed sliding dovetail" for this](https://www.finewoodworking.com/1997/04/01/housed-sliding-dovetails) and, best of all, they have a term for the open insertion area: *escapement*. How is `escapement_length` for the term? (I'll also fix the rest of the language)

adrianVmariano:
When I said that you would see the entry hole I was thinking about a practical woodworking joint.  The joint you're making here doesn't appear to be a practical woodworking joint.  I guess it depends on how hard you want to work...or if you're cutting it with a CNC.  The Bracht video is interesting---his focus is on Japanese joinery, and it makes sense that any wacky joint you dream up appears somewhere on a list of Japanese joints.   But also, this joint can only be cut in wood on the end of a board (or it will be weak).  It's normally as you note, used to install shelves and the joint is exposed from the back.  Making it blind requires cutting off half the joint, which adds a lot of work, and also really to me, "fully housed" suggests a joint without a hole on one side---not even a hidden hole.  

You attributed the term to AI in the message.  I think that the Fine Woodworking article is presenting an exotic method.  I didn't give them my email to read the article---are the dovetail keys screwed on to the case?   The term "escapement" seems to have been grabbed by the author to refer to this but I can't find any indication that it normally has that meaning.  (The standard meaning is a mechanism in a clock.)   Even if there is some obscure reference, perhaps ancient historical use for "escapement" to mean the entry hole, it's obscure, and hence won't be meaningful to users.  So it's not a good term.

adrianVmariano:
As noted above, "escapement" is a bad term.  Nobody will no what it means (no intuition) and you can't even look it up in a dictionary, since the meaning you want seems to be unique to popular woodworking.   The ideal name is something users can guess the meaning of when they see it, and have a chance of remembering if they read the docs last month.   By those measures,  "entry_hole_length" is definitely better, but it's rather long.

coryrc:
Regardless of anyone's opinion of its practicality using wood, I just used it for wall mounting a 3d print.

I will change to entry_hole_length when I'm back to the computer. I'm not excited about "hole" though, but it ultimately isn't that important.

adrianVmariano:
Oh, I never said this joint wasn't a useful 3d printed joint.  It's just that the language around dovetails comes from woodworking and for woodworking, this is a weird and difficult joint.  

I agree that "hole" doesn't seem ideal.  Is "entry_slot_length" better?

coryrc:
That's a definite improvement.

PR #1751, opened 2025-07-17
Fix FPT: Internal threads can't have blunt start

coryrc:
You can't start the thread unless the opening extends all the way to the end...

Maybe this should be enforced in generic_threaded_rod?


adrianVmariano:
Not sure what you mean.  Of course internal threads can have blunt start.  That is the default, in fact.

coryrc:
Okay, this is what I get for only looking at the mask and not applying it to the object first. I confused the outer and inner portions.

PR #1771, opened 2025-08-11
Add error message with solution to skin

coryrc:
Maybe you're refactoring a bunch of nested arguments to skin and forget the first argument is supposed to be a list and you wasted an embarrassing amount of time figuring it out :) In this case, the error message changes from

Profiles [0, 1, 2, 3, 4, 5, 6, 7, 8] are not a paths or have length less than 3.

to

The first argument to `skin` must be a list of paths


amatulic:
That seems like an over-complicated way to detect a list of paths. Instead of
````
assert(is_list(profiles)&&is_list(profiles[0])&&is_list(profiles[0][0]), "\nThe first argument to `skin` must be a list of paths")
````
Is there any reason why you can't use this?
````
assert(is_list(profiles) && is_path(profiles[0]), "\nThe first argument to `skin` must be a list of paths")
````

coryrc:
Done

PR #1781, opened 2025-08-23
Fixed octa spheroid() symmetry.

revarbat:
The octagon style of spheroid() was malformed and asymmetrical due to the use of spherical_to_xyz().  This code fixes that.

### DIFF
--- a/hinges.scad
+++ b/hinges.scad
@@ -458,7 +458,7 @@ module _knuckle_hinge_profile(offset, arm_height, arm_angle=45, knuckle_diam=4, 
         if (is_num(clear_top)) left(.1)fwd(clearance) rect([.1+clear_top, knuckle_diam+1+clearance], anchor=BOT+LEFT);
         if (is_def(clip)) fwd(clip) left(.1) rect([offset+knuckle_diam, ofs+round_bot+knuckle_diam+abs(clip)],anchor=BACK+LEFT);
       }
-      right(offset)ellipse(d=knuckle_diam,realign=true,circum=true);
+      right(offset)ellipse(d=knuckle_diam, circum=true);
     }
     if (is_num(pin_diam) && pin_diam>0){
       $fn = default(pin_fn,$fn);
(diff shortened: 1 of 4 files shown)


adrianVmariano:
Should we wait for the dust to settle on the OpenSCAD PR so that they match?   I haven't been paying close attention to the discussion about the optimization of the sphere.  

Also I wonder (haven't looked) how the "octa" sphere works with circum=true and whether it just calls the regular octasphere or also needs updating.

revarbat:
The circum=true octa sphere basically does `hull(_dual_vertices(spheroid(r=r, style="octa",circum=false)))`, so no update needed for that.

revarbat:
As for waiting for the final builtin version, that could take years, and the existing implementation in BOSL2 is definitely a bug.  This code fixes the asymmetries, even if it is not pixel perfect to the eventual design.  We can correct it for an exact match if and when it actually becomes an available thing.

revarbat:
Also, based on the statistics, this code is slightly better than the proposals so far.

revarbat:
ECHO: "new octa:"
ECHO: "  Face count: 128"
ECHO: "  Radius: 1"
ECHO: "  Face Area min: 0.0760653"
ECHO: "  Face Area max: 0.110289"
ECHO: "  Face Area avg: 0.0935639"
ECHO: "  Face Area rms: 0.0119409"
ECHO: "  Face Area cmc: 0.0129193"
ECHO: "  Face Tension min: 0.0337573"
ECHO: "  Face Tension max: 0.0436897"
ECHO: "  Face Tension avg: 0.0384038"
ECHO: "  Face Tension rms: 0.00356386"
ECHO: "  Face Tension cmc: 0.0038193"
ECHO: "  Edge Length min: 0.390181"
ECHO: "  Edge Length max: 0.541196"
ECHO: "  Edge Length avg: 0.472324"
ECHO: "  Edge Length rms: 0.0535752"
ECHO: "  Edge Length cmc: 0.0592344"
ECHO: "  Vertex Radii min: 1"
ECHO: "  Vertex Radii max: 1"
ECHO: "  Vertex Radii avg: 1"
ECHO: "  Vertex Radii rms: 4.05396e-17"
ECHO: "  Vertex Radii cmc: 5.67183e-17"

revarbat:
BTW, I've checked the icosa sphere, and it does not appear to have the same flaw.

adrianVmariano:
I think that the PR will settle.  If you have something better than what's being talked in the PR we should post it there, probably.  I'm not saying wait for it to appear in OpenSCAD, but wait for the PR to solidify and then we should match the PR.

PR #1783, opened 2025-08-23
Add Grip Fins to joiners

coryrc:
Grip fins allow a cylindrical object (or even ovoid-cross-section) to be held tightly, but also removable and reusable.

My searching didn't find anybody else having made these in SCAD. 

In [this old Slant 3D video](https://www.youtube.com/watch?v=yzg_NXM-NRs), he shows straight fins used. This is the one implemented so far.

In [this newer Slant 3D video](https://youtu.be/Bd7Yyn61XWQ?t=371) he uses a different shape of fin that I haven't implemented yet -- at one point I tried, but they don't work at small scale. Judging from the video it's theoretically as easy as taking the one I made and diffing it out.

This is the version I've used to make grip fins for holding PTFE tubes for my 3d printer. They aren't attachable yet and the docs aren't finished and the naming is dumb and the tests aren't in the right place, which is why it's draft.

### DIFF
--- a/joiners.scad
+++ b/joiners.scad
@@ -1439,4 +1439,110 @@ module hirth(n, ir, or, id, od, tooth_angle=60, cone_angle=0, chamfer, rounding,
   }
 }
 
+// Asserts a solution exists when called
+function _dist2inner_for_cylindrical_clamp(angle_from_tangent, or, ir) =
+    is_undef(angle_from_tangent)?sqrt(pow(or,2)-pow(ir,2)):
+    assert(angle_from_tangent<90)
+    assert(angle_from_tangent>=0)
+    let(
+        // The line leaving the circumference is the tangent line of some concentric circle with radius:
+        tangent_circle_radius = or * sin(angle_from_tangent),
+        length_from_starting_point_to_tangent_circle = or * cos(angle_from_tangent)
+    )
+    assert(tangent_circle_radius<=ir, "The angle given for cylindrical clamp fins does not allow it to intersect with the inner circle radius")
+    let(
+        length_from_tangent_circle_tangent_to_intersection_of_inner_circle = sqrt(pow(ir,2)-pow(tangent_circle_radius,2)),
+        fin_length=length_from_starting_point_to_tangent_circle-length_from_tangent_circle_tangent_to_intersection_of_inner_circle
+    )
+    echo(tangent_circle_radius=tangent_circle_radius,length_from_starting_point_to_tangent_circle=length_from_starting_point_to_tangent_circle,length_from_tangent_circle_tangent_to_intersection_of_inner_circle=length_from_tangent_circle_tangent_to_intersection_of_inner_circle,fin_length=fin_length)
+    fin_length;
+
+assert(_dist2inner_for_cylindrical_clamp(0,8,5)==3);
+
+// Module: cylindrical_clamp()
+// Synopsis: Builds a finned clamp for circular objects.
+// SynTags: Geom
+// Topics: Joiners, Parts
+// Usage:
+//   cylindrical_clamp(h, od, id, [fin_thick=0.4], [num=3], [fin_angle=30]);
+//   cylindrical_clamp(h, od1=, od2=, id1=, id2=, ...);
+//   cylindrical_clamp(h, od, id, [shift=[dx,dy]], ...);
+// Description:
+//   Generate a clamp designed to hold a cylindrical or slightly-conical object.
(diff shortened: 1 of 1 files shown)


amatulic:
The second version is preferable to the first, but the difference is trivial. In the first video, he makes fins protruding from the side of the hole. In the second version, he makes fins by cutting slots into the side of a smaller hole.

The second version is far, far better because the resulting fins taper, distributing the flexing loads more evenly across the span of the fin. If the fins are constant thickness like in the first video, the flexing load gets concentrated at the base of each fin, which increases the likelihood of fatigue and breaking. Or in the case of PETG, the plastic doesn't break but deforms permanently, eliminating the friction-fit feature that was intended in the first place.

Standard design guidance for flexing snap-connectors also advise that the base of the flexing part should be double the thickness of the tip, to reduce fatigue and the chance of failure.

Therefore, I would abandon the first idea and focus on the second.

coryrc:
> Standard design guidance for flexing snap-connectors also advise that the base of the flexing part should be double the thickness of the tip, to reduce fatigue and the chance of failure.

Nice! I'll use that as the default. I assume optimal is around 2x and not "at least 2x"?

With the size I was using, it's actually that the tip is twice the size it needs to be because a single extrusion did not hold well, so it needed to be at least 2x the line width (or, with arachne, 1.5x? I wrote 1.75x but I think it goes down to 75% line width so that also needs confirming).

amatulic:
The guidance I've seen is to taper the cross-section 2X from base to tip. It doesn't have to be exact. A higher ratio moves the stress toward the tip, and a lower ratio moves it toward the base. Often I've had to fudge it due to space constraints, going less than 2X.

See https://fab.cba.mit.edu/classes/S62.12/people/vernelle.noel/Plastic_Snap_fit_design.pdf pages 8-9 -- that isn't the design guide I originally found, but it says basically the same thing.

Some of my own designs where I've used this concept:
* [Pullcopter](https://www.printables.com/model/160112-way-over-engineered-pullcopter-v3) -- one of my most popular designs. The handle is two parts that snap together. There is a long flexing piece in the main handle (it's against the build plate as it's printed) and it snaps into a latch in the top part. If you print this, print a few extra propellers. I have lost a propeller every time I fly it. They seem attracted to trees and roofs of buildings and indoor ceiling rafters.
* [JST connector pliers](https://www.printables.com/model/459462-over-engineered-jst-connector-pliers) - in this case the spring leaves are tapered in thickness, and optimally printed horizontally, as a flexing part should be. I haven't noticed any fatigue issues with PETG or PLA, likely due to the spring leaves being long compared to the amount of flex.
* [Cat feeder frame](https://www.printables.com/model/892827-surefeed-microchip-pet-bowl-frame-more-reinforced) - the snap fit thickness gets wider perpendicular to the flex direction rather than thicker in the same direction as the flex. Printing a flexing part vertically like this is sub-optimal because the stress is on layer boundaries, which is the weakest feature of any 3D print.
* [Replacement IKEA lunch box clasp](https://www.printables.com/model/599359-ikea-flottig-lunch-box-clasp-replacement) - same principle, but I wanted to maximize the contact area of the end, so I just accepted whatever smaller ratio resulted from a 45° slope. This also is suboptimal-printed vertically.

adrianVmariano:
What's the status of this?  Is it going to move forward?

coryrc:
> What's the status of this? Is it going to move forward?

Yes, just not my biggest priority right now, unless it's causing anybody problems.

adrianVmariano:
I was just trying to clean up the PRs.  Ideally there would be zero, but if you're planning to return to it then we can leave it for now.

revarbat:
The TODO: lines were confusing docsgen and causing checks to fail, so I tweaked them to not do that.

revarbat:
Okay, this is technically mergable now, but I have not yet actually run the code yet.

amatulic:
It shouldn't be merged until the code includes at least one example.

adrianVmariano:
I thought this was very much "in development" and not ready for use right now.

amatulic:
Yes, well, I made that comment with an abundance of caution, remembering when you were on vacation and my first draft of isosurfaces+metaballs went live before it was ready. This grip fin feature is a fairly simple thing and should be much easier to get into shape for merging.

adrianVmariano:
I'm not sure which comment you mean.  I thought @coryrc wasn't finished working on this and it seems like there were a bunch of suggestions that I had the impression were as yet not implemented.

coryrc:
> I'm not sure which comment you mean. I thought @coryrc wasn't finished working on this and it seems like there were a bunch of suggestions that I had the impression were as yet not implemented.

That's correct, I really liked @amatulic suggestions. It works fine as-is (or did, at least), I just haven't gotten around to it yet, getting stuck on some other projects. I actually do have something coming up that would be fun to use these with.

PR #1793, opened 2025-09-05
Implement a zip() function to iterate over lists in parallel

pipatron:
I missed a `zip()` function in BOSL2 (at least I couldn't find one), and implemented one as an exercise in writing test code and documentation.

`zip(a,b)` takes items from lists _a_ and _b_ in parallel and can be used for things such as processing a path and its tangents. `zip(a,b) == [ [a[0],b[0]], [a[1],b[1]], ... ]`

The weird name _zip_ comes from the same function in python, perl, C++ and others. I don't like it, but that's common usage now.

The already existing `transpose()` can be used to do this for some lists, but is really tailored for matrices and doesn't like empty lists and uneven lengths etc.

### DIFF
--- a/lists.scad
+++ b/lists.scad
@@ -1074,6 +1074,30 @@ function full_flatten(l) =
     [for (a=l) if (is_list(a)) (each full_flatten(a)) else a];
 
 
+// Function: zip()
+// Synopsis: Group elements from multiple lists by index.
+// Topics: List Handling, Iteration
+// See Also: pair(), triplet(), transpose()
+// Usage:
+//   list = zip(lists);
+//   list = zip(LIST1,LIST2);
+// Description:
+//   Returns a list composed of elements from the given lists. Each entry in the new list is composed of elements at that corresponding index from each input. If any list run out of elements before others, "undef" will be used.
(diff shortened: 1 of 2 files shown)


revarbat:
I'm fairly certain that I wrote an implementation of this a few years back, but it was deemed not terribly useful, given that:

    for (c = zip(a,b)) foo(c[0],c[1])

Is actually more verbose than:

    for (i = idx(a)) foo(a[i],b[i])

revarbat:
Also, I think you can do this with `hstack()`.

adrianVmariano:
We definitely had zip and found it basically not useful.  In python you can do

```
for path,tagent in zip(pathlist,tanglist):
```

but what happens in OpenSCAD looks like

`for(entry=zip(pathlist,tanglist))    do_something(entry[0])+something_else(entry[1])`

and not only is it not particularly parsimonious of character count, the code is actually worse, harder to understand, than the direct indexing

```
for(i=idx(pathlist)) do_something(pathlist[i]) + something_else(tangent[i])
```

I think zip sort of lives on in hstack, but it doesn't do quite the same thing because it concatenates the items instead of just combining them in a list.

I don't see a compelling reason at the moment to revisit our decision to eliminate zip().

PR #1801, opened 2025-09-15
Allow unnamed anchors to be aliased

snoack:
This will enable aliases like this which currently don't work:

```
cuboid(20)
  change_anchors(alias=[["slot", TOP]])
    show_anchors();
```


adrianVmariano:
I had to change the docs to reflect this change and also I added some error checking.  It was easier for me to do this in a separate PR.

PR #1851, opened 2025-12-01
extra rabbit clip configurations

pelletjl:
add a half-lock configuration to rabbit clips and add extra configuration to double-sided pins

### DIFF
--- a/joiners.scad
+++ b/joiners.scad
@@ -1152,7 +1152,8 @@ module rabbit_clip(type, length, width,  snap, thickness, depth, compression=0.1
     assert(is_num(depth) && depth>0, "Depth must be a positive value")
     assert(is_num(compression) && compression >= 0, "Compression must be a nonnegative value")
     assert(is_num(lock_clearance))
-    assert(valid_lock(), "Invalid lock value")
+    assert(valid_lock(), type != "double" ? str("lock must be either true, false, LEFT, or RIGHT for type \"", type, "\"")
+            : "for type \"double\", lock must be either true, false, TOP, BOTTOM, LEFT, RIGHT, a corner (e.g. TOP+LEFT), or a vector of edges and/or corners")
     assert(in_list(type,legal_types),str("type must be one of ",legal_types));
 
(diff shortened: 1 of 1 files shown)


adrianVmariano:
Would it be cleaner to have lock= take string parameters that specify the different lock states instead of introducing lots of booleans that then are potentially conflicting?  e.g. lock=true/false works as currently, lock="half" gives you the half lock.  I'm not sure what to do about the double case.  There are a total of 8 lock states, since there are four locations that may or may not lock.  Another idea instead of strings would be a boolean list, like lock=[true,false,false,true] to specify the desired lock state for each position.  It would be only length 2 for single-ended pins.

pelletjl:
I wasn't sure if mixing argument types was an acceptable practice, and I was trying to preserve backward compatibility. I originally allowed conflicts to pass through, but documenting which ones took precedence seemed a bit much. I can try to clean it up if mixing types isn't an issue.

adrianVmariano:
We often mix argument types.  Allowing either a single boolean or a list of booleans is actually a very common strategy where a single false is equivalent to a list of false values (e.g. apply force_list to turn singleton into a list).  That's the simplest solution,  I'm not sure if it makes sense to have the order go clockwise around the pin or if it should be left right and then left right again for the second pin if present.  There's also the question of whether to expand [a,b] into [a,a,b,b] for a double pin, for example when you want one end to lock and the other end to not lock.  I think using string arguments here is messier and not necessary.

pelletjl:
what about, in addition to the current true/false for all edges, allowing LEFT or RIGHT, which would apply to all types as well as TOP or BOTTOM for "double". Also with "double" we could accept a corner or array of corners to lock, e.g. lock=[TOP+LEFT, BOTTOM+RIGHT] or lock=TOP+LEFT?

adrianVmariano:
I think the proposal above is reasonable, using the direction vectors.  When the input is invalid the error message should probably tell the user what options are valid for the current configuration,

pelletjl:
Done!

PR #1881, opened 2026-01-14
fix for knuckle_hinge using both in_place and knuckle_clearance

### LINKED ISSUE #1878: [BUG] knuckle_hinge: knuckle_clearance prevents in_place hinges from working
When working with a knuckle_hinge, specifying a knuckle_clearance removes the interlocking pins of an in_place hinge.

Example:
(derived from examples)
```
include <BOSL2/std.scad>
include <BOSL2/hinges.scad>
$fn=32;
thickness=2;
clearance=0.2;
zrot_copies([0,180])
  color(["green","gold"][$idx])
  diff()
    xrot(180*$idx, cp=[0, 0, 7/2+thickness*2/3 + clearance])
    fwd(clearance/2)
      cuboid([20,thickness+1,7],anchor=BACK) {
        down(thickness/3)
        position(TOP+BACK)
          knuckle_hinge(20, segs=5, offset=thickness+clearance,
                        inner=$idx==0,
                        knuckle_clearance=clearance,  // comment out this line to see the in_place pins
                        clearance=clearance/2, arm_angle=90,
                        knuckle_diam=2*thickness+clearance,
                        clear_top=true, in_place=true);
        attach(TOP, BOTTOM, align=FRONT)
            cuboid([20, thickness, thickness-2.5*clearance]);
                        }
```

pelletjl:
fixes #1878 for default (untagged) case

### DIFF
--- a/hinges.scad
+++ b/hinges.scad
@@ -387,16 +387,7 @@ module knuckle_hinge(length, segs, offset, inner=false, arm_height=0, arm_angle=
         union(){}
     }
   }
-  module in_place_clearance_tag() {
-      // don't set the tag if it's already set for backward compatibility
-      if(in_place != false && knuckle_clearance && $tag == "") {
-          tag("keep")
-              children();
-      } else {
-          children();
-      }
-  }
-  in_place_clearance_tag()
+  default_tag("keep", do_tag=in_place != false && knuckle_clearance)
   attachable(anchor,spin,orient,
              size=[length,
                    arm_height+offset/tan(arm_angle)+knuckle_diam/2+knuckle_diam/2/sin(arm_angle),


adrianVmariano:
Can't you use `default_tag()` to do this?

pelletjl:
I was trying to narrow the scope of when the default tag is applied just in case there were any unknown side effects, but I can try using `default_tag()` instead.

pelletjl:
Ah, I see how to use `default_tag()` to do the same thing now. The documentation for it wasn't clear.

adrianVmariano:
The hinge still has a problem as shown by top_half(z=4.8):

image

Some extra material is connecting the two sides of the hinge together.

pelletjl:
That part is from the extra backstop that I added. It goes away without the second (attached) child, which I've adjusted in my design.

adrianVmariano:
Would it suffice to clarify default_tag if I changed the description to "conditionally set a default tag"?

pelletjl:
The description was fine, I just didn't see the argument in the module signature (in the `usage` section) even though it was in the argument descriptions. But if you're in there doing other things and wanted to update the description, it wouldn't hurt.

adrianVmariano:
I can update the docs and the update will merge someday.

PR #1891, opened 2026-01-29
Fix trigonometry alias problem

### LINKED ISSUE #1805: [BUG]
`hyp_adj_to_opp()` has an alias `adj_hyp_to_opp()`. They should NOT be aliases, as they have different argument order.  There are several related functions in trigonometry.scad with similar problems.

RAMilewski:
Moved former "aliases" to full function entries.

Resolves #1805

### DIFF
--- a/trigonometry.scad
+++ b/trigonometry.scad
@@ -25,8 +25,8 @@ _BOSL2_TRIGONOMETRY = is_undef(_BOSL2_STD) && (is_undef(BOSL2_NO_STD_WARNING) ||
 //   c = law_of_cosines(a, b, C=);
 // Description:
 //   Applies the Law of Cosines for an arbitrary triangle.  Given three side lengths, returns the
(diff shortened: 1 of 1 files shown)


adrianVmariano:
Maybe opp_hyp_to_adj the synopsis should flip order to match

The terms "primary angle" and "primary corner" appear throughout.  There were there before, so not your fault....but I've never heard them and neither has the Claude AI, or google.  So maybe not a term to use...or not without definition, at least.   (And "primary" suggests biggest, maybe, which isn't right.)  

Maybe "reference angle"?  I came up with that on my own, but Claude also suggested it.  

Instead of:

     ang = The angle in degrees of the primary corner of the right triangle.

maybe just

    ang = reference angle of the right triangle in degrees

in all the places where it appears?  

ang_adj_to_opp()

That's all I noticed.

PR #1939, opened 2026-03-15
Include install options for users running Fedora Linux

hegjon:
Fedora Linux provides easy to install and upgrade BOSL2 for their users by providing a RPM package in their official repository

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,8 @@ Users on Fedora Linux can install directly from their package repository by runn
 
 ```sudo dnf install openscad-bosl2```
 
+ - NOTE: The version might be older than the current release.
+
 ## Examples
 A lot of the features of this library are to allow shorter, easier-to-read, intent-based coding.  For example:
 


adrianVmariano:
How up-to-date is the fedora version of BOSL2?  Should there be a note that it may be an older version?

hegjon:
Disclaimer: I am the maintainer for the package at Fedora Linux.

Yes, adding a note that the version might lag behind.

Since there have not been a release/tag before recently, then the updates have been random. With tag/release, it is easy to monitoring for new releases.


Here is the update history:
https://bodhi.fedoraproject.org/updates/?packages=openscad-bosl2

adrianVmariano:
Did you add the note about possible version lag?  I don't see a new commit.

hegjon:
Added the note now

hegjon:
As a side note; I have added BOSL2 to https://release-monitoring.org/project/389249/

PR #1963, opened 2026-04-05
feat: add write() module for advanced 3D text with auto-sizing and wrapping (#1642)

junhinhow:
## Summary
New `write.scad` module (not included in std.scad, requires separate include):

- **`write()`** module: advanced 3D text with:
  - **Auto-sizing**: omit `font_size` + provide `box`/`width`/`height` → text scales to fit
  - **Word wrapping**: provide `font_size` + `width` → automatic line breaks at word boundaries
  - **Multi-line**: accepts array of strings or string with `\n` newlines
  - **Alignment**: horizontal (`left`/`center`/`right`) and vertical (`top`/`center`/`bottom`/`baseline`)
  - **Margins**: configurable padding inside the bounding box
  - Full `attachable()` integration with anchor/spin/orient

- **`get_font_size()`** function: converts between typographic units
  - `font_size` (traditional point-like size)
  - `em_size` (em square / font design box)
  - `cap_height` (capital letter height, with configurable ratio)
  - Automatically handles the OpenSCAD 0.72 size factor

Usage: `include <BOSL2/std.scad>` then `include <BOSL2/write.scad>`

## Test plan
- [x] `get_font_size()` returns correct values for all 3 unit types
- [x] Basic text renders valid 3D geometry
- [x] Multi-line text (array input) renders correctly
- [x] Auto-sizing fits text within specified box
- [x] Word wrapping breaks lines at word boundaries
- [x] All alignment combinations work (left/center/right × top/center/bottom)
- [x] Attachment to parent objects works correctly

Addresses #1642

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/test_write.scadtest
+++ b/tests/test_write.scadtest
@@ -1,0 +1,120 @@
+[[test]]
+name = "test_get_font_size"
+script = '''
+include <../std.scad>
+include <../write.scad>
+
+module test_get_font_size() {
+    // font_size: divide por 0.72 para obter OpenSCAD size
+    sz1 = get_font_size(font_size=12);
+    assert_approx(sz1, 12/0.72);
+
+    // em_size: multiplica por 0.72
+    sz2 = get_font_size(em_size=10);
+    assert_approx(sz2, 10*0.72);
+
+    // cap_height: cap / ratio / 0.72
+    sz3 = get_font_size(cap_height=5);
+    assert_approx(sz3, 5/0.7/0.72);
+
+    // cap_height com ratio customizado
+    sz4 = get_font_size(cap_height=5, cap_ratio=0.65);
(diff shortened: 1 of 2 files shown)


adrianVmariano:
Presumably we don't want to use this given that @amatulic has mostly written a write() module.

amatulic:
Doesn't address all of #1642 but looks interesting. I never tried Claude for coding.

The wrapping doesn't appear to handle proportional fonts. 

I need to finish this. It's nearly all done (auto-size wrapping to fit a vertical bound still not tested, and the anchoring stuff is not there yet). Right now I'm busy trying to update a model for a $500 contest on Thingiverse, which will take me until Monday, and after that I should be able to work on the write() module.

Issue #521, opened 2021-04-25
vnf_vertex_array styles: when to use which ones (examples), quincunx and concave


adrianVmariano:
I'm wondering about the styles for vnf_vertex array.  Are they all useful?  And if so, should we add a concave style?  I think we need some examples showing cases where the different VNF styles work best, and how the wrong choice looks bad.  

Is there ever a situation where the quincunx style produces a good (desirable) result?   Seeing the example with gears where it made things look bad made me aware of this issue: it introduces a new point which is only actually on the polyhedron if the face being split is coplanar.   So it seems like we have a situation where if the face is coplanar this is a good choice....but it won't look better than the alternative.  If the face is nearly coplanar it should also be a decent choice, but it probably still won't look better.  And if the face is far from coplanar it's going to twist the shape in a weird way and the result will be bad.   I'm inclined to think that we should remove this style.  

A similar question arises for the convex style.  Is there some case where it produces the best result and is different than the other results?  Presumably this would be for convex shapes? If such an example exists then presumably it can be inverted to require the concave style, so then presumably the concave style should be implemented as well.

So I'm thinking that "quincunx" should go away.  And either we add "concave" or we also get rid of "convex".  

Whichever styles we keep, there needs to be some examples in the docs showing how the style differ and when the various styles look good/bad.

revarbat:
"quincunx" was added for the case where the surface flexes both ways, making both the default and "alt" styles incorrect for parts of the same surface. Consider passing `function (x,y) sin(x/3) * sin(y/3)` to `heightfield()`. You get places on the surface where it naturally wants to bend against how it gets triangulated. 

It preceded "min_edge" and "convex". I'm fairly sure "convex" doesn't make a better look for this case. Nor would a theoretical "concave", which we maybe should implement. Dunno about "min_edge".

adrianVmariano:
I'm having a hard time figuring out whether quincunx actually looks better for that case.  The problem is that it's hard to know how to make a good comparison.  I tried doing difference of a high resolution version of your height field and differencing with low res using different styles.  By this measure, quincunx performed worse than everything except convex.  

[...]

PR #1967, opened 2026-04-06
docs: add style selection guide and visual comparison for vnf_vertex_array (#521)

junhinhow:
## Summary
- Add practical guidance on choosing triangulation styles for vnf_vertex_array()
- Add visual comparison example showing 5 styles side-by-side on the same wavy surface with wireframe overlay
- Document when to use each style: default/alt for smooth, min_edge for varying curvature, convex/concave for textures, quincunx for heightfields (with saddle artifact warning), quad for speed

## Test plan
- [x] New example compiles in OpenSCAD
- [x] Documentation follows BOSL2 format conventions
- [x] No changes to code logic

🤖 Generated with [Claude Code](https://claude.com/claude-code)


adrianVmariano:
There are ten supported styles now in vnf_vertex_array, so if this only shows 5 it's incomplete.  

It also just adds one example.  This topic probably needs 15 examples to identify how the styles interact with the data and when you should choose which one.

PR #1992, opened 2026-06-03
Reworked examples in NURBS Interpolation.

RAMilewski:


nurbs.scad
```diff
@@ -45,7 +45,7 @@ _BOSL2_NURBS = is_undef(_BOSL2_STD) && (is_undef(BOSL2_NO_STD_WARNING) || !BOSL2
 //   number of control points.  A single NURBS is more like a bezier **path** than like a single bezier spline.
 //   .
 //   A NURBS or B-spline is a curve made from a moving average of several Bezier curves.  The knots specify when one Bezier fades
-//   away to be replaced by the next one.  The knot list is a non-increasing list of values that you specify using two parameters,
+//   away to be replaced by the next one.  The knot list is a non-decreasing list of values that you specify using two parameters,
```
duncanbeevers, line 48: This might read clearer as `monotonically increasing`.

adrianVmariano: I actually think it's less clear, since it requires the reader to know the precise math jargon meaning of "monotonically increasing", and it's really crucial that repeats are permitted.

PR #2020, opened 2026-08-21
Add starts parameter to generic_bottle_cap() for multi-start threads

HannesGitH:
Plastic bottle necks — especially the larger PET finishes (38 mm and up) — commonly use 2- or 3-start threads, but `generic_bottle_cap()` could only produce single-start threads. I ran into this making an adapter cap for a 38 mm PET bottle: the pitch was right, but the cap only engaged for a fraction of a turn before binding, because the neck lead is `2 * pitch`.

`thread_helix()` already supports `starts`, so this just passes it through:

- New `starts` argument (default 1), documented in the Arguments block and the Description.
- `turns` is divided by `starts`, since `thread_helix()` rises `pitch * starts` per revolution — without that the thread would overflow the cap cavity.
- `pitch` keeps its existing meaning (axial distance between adjacent threads, i.e. what you measure with calipers on the neck), so the lead is `pitch * starts`.
- One added doc example showing a 2-start cap.

Backwards compatible: with the default `starts=1` the generated geometry is unchanged — I rendered `generic_bottle_cap(texture="knurled",neck_od=25,thread_od=30,height=10)` before and after and the STLs are byte-identical (md5 `2a525bc9150080fda6e2a76673f74858`).

The 2- and 3-start variants render manifold with no warnings under `--hardwarnings`.

Happy to add the matching `starts` to `generic_bottle_neck()` in a follow-up (or here, if you prefer) — I kept this one minimal.

### DIFF
--- a/bottlecaps.scad
+++ b/bottlecaps.scad
@@ -551,6 +551,10 @@ function generic_bottle_neck(
 // Description:
 //   Creates a basic threaded cap given specifications. You must give exactly two of `thread_od`, `neck_od` and `thread_depth` to
 //   specify the thread geometry. Most glass bottles conform to the SPI standard and caps for them may be more easily produced using {{sp_cap()}}.
+//   .
+//   Plastic bottle necks often use multi-start threads, which you can create by setting `starts` to the number of
+//   thread starts. The starts are spaced evenly around the cap, and `pitch` remains the axial distance between
+//   adjacent threads, so the lead (the axial travel of one revolution) is `pitch * starts`.
 // Arguments:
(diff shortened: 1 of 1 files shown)


amatulic:
I agree it's better to have matching caps and necks. The changes to the file look reasonable and minimal, and may be potentially useful.

I can't say I've seen a bottle cap with multiple starts, but I've certainly seen wide-mouth things like jars with multiple-start necks. These typically have metal lids with multiple lugs that catch the threads. In that sense, having a bottle neck model may be more useful than a bottle cap.

PR #2030, opened 2026-09-02
added str_replace(), improved speed of str_find()

amatulic:
Changes to strings.scad:
* Corrected typo in one of the `str_find()` examples
* ~Replaced `str_find()` with more efficient version, tested on the given examples to be 4X faster, and other tests showed up to 7X improvement~
* (Update to PR) Replaced the `_str_find_all()` private function of `str_find()` with more efficient version.
* Added new function `str_replace()` ~, which is significantly slower than `str_replace_char()` (and noted this in the docs)~ (tested \~10X faster than earlier version in this PR)
* Corrected error in output for example in `str_replace_char()`


amatulic:
@adrianVmariano -- because my version of `str_find()` was advantageous only in the case when `all=true`, I have replaced only the internal function `_str_find_all()` with a faster version, and left the rest of `str_find()` alone.

amatulic:
@adrianVmariano - I have also replaced my original `str_replace()` with a new one that is 10X faster, now about the same speed as `str_replace_char()`. It doesn't do any string manipulation or calls to substr(), instead it builds out a list of unicode values using `each` and then combining them all into a string using `chr()`. Examples I ran 1000 times took about 300 ms versus 3.5 sec for the old version.

PR #2048, opened 2026-09-12
nurbs_sheet() fixes

RAMilewski:


RAMilewski:
This doesn't include the code for Adrian's suggestion to add caps, cap1 and cap2 to nurbs_sheet().   I fought with Claude for about 15 rounds and lost.  I'll try again with fable next time, but wanted to get all the other changes in.

This is the best Claude could do:

Screenshot 2026-09-13 at 1 15 09 PM

revarbat:
The error is correct — it's just describing a file @RAMilewski never sees.

CI checks out the PR merge ref, not the branch head. On the branch there's one nurbs_length() at line 1390, which is why the reported lines 1441 and 1742 look like nonsense. In the merge of #2048 into current master, both of those lines are exactly // Function: nurbs_length():

merge cfd3b3d  nurbs.scad
  1441: // Function: nurbs_length()     ← from master
  1465: function nurbs_length(control, degree, u=[0,1], mult, weights, type="clamped", knots)
  1742: // Function: nurbs_length()     ← from this PR
  1770: function nurbs_length(control, degree, mult, weights, type="clamped", knots,
                              start_u=0, end_u=1, tol=1e-6, maxdepth=10)

master gained its own nurbs_length() in 19a2ce92 "Add nurbs_length and nurbs_find_u", after this PR branched. The two live in different regions of the file, so git merged both additions without a conflict — a clean merge that is nevertheless a duplicate declaration.

They're also genuinely different functions: master's takes start_u/end_u/tol/maxdepth and does adaptive subdivision; the PR's takes u=[0,1] and does Gauss-Legendre quadrature. So it's a real collision needing a decision, not a mechanical fix.

What @RAMilewski needs to do: merge current master into the branch, then keep one implementation — either drop theirs in favour of master's, or replace master's and reconcile the signature. CI will stay red until then, and nothing about their branch alone will show why.

I reproduced the negative locally too: belfryscad --docsgen -Tmf on the PR head passes clean, which confirms it's the merge and not our docsgen.

Worth noting the error message could say so — "at nurbs.scad:1742" with no hint that this is the merge result is what made it look absurd. If you want, I can make docsgen's redeclaration error print both declarations' surrounding context, or have BOSL2's CheckDocs note it's testing the merge.

